PR #39, opened 2022-07-01
alpha 2 fixes

Aprillion:


app/components/question.tsx
```diff
@@ -35,6 +34,31 @@ export default function Question({
...
+        hideEl?.addEventListener('click', hideLong)
+
+        return () => {
+          showEl.removeEventListener('click', showLong)
+          hideEl?.removeEventListener('click', hideLong)
+        }
+      }
+    }
+  }, [isExpandedAfterLoading])
+
```
ccstan99, line 61: Is this something that's marked in the wiki with short/long description?

ccstan99 (approved):
Looks good. FYI, this tooltip shows up when the search box is empty.
Screen Shot 2022-07-01 at 11 44 19 AM

app/components/question.tsx
Aprillion: yup, e.g. https://stampy.ai/wiki/How_do_I_format_answers_on_Stampy%3F => https://ui.stampy.ai/?state=7503_

Aprillion: I would be more happy with just 1 length of each answer, but @plexish is still confident this is a useful feature ¯\\\_(ツ)_/¯

Aprillion:
> Looks good. FYI, this tooltip shows up when the search box is empty.

hm.. I don't see it in the HTML or in my browser .. could it be from some browser extension? if it happens with some particular repro steps, please open a new issue and we can investigate later

app/components/question.tsx
ccstan99: Yes, I personally believe it's a healthy exercise to give concise answers in an FAQ. If an answer is really long, it probably really should just be broken up into related questions/answers.

PR #40, opened 2022-07-01
update icon & gen codings

ccstan99:
Updated icon and preview images with current logo/design. 
Only generate encodings if number of questions changed. Currently, things actually work but generate an error, so the final action fails. I've left placeholder to check things. Within GitHub actions, there doesn't seem to be a quick way to check if files have changed. Need to do a bit of checking/processing of output. Someone can figure this out this later. I think I better look at the styling for the launch 7/5.


questions-encodings.mjs
```diff
@@ -1,30 +1,45 @@
...
+console.log(`${response.length} questions fetched from the stampy.ai wiki.`)
 
 const questions = response.map(({fulltext}) => fulltext)
 const questionsNormalized = questions.map((question) => question.toLowerCase())
+const numQs = questions.length
+
+if (prevNumQs == numQs) {
+  console.log(`Number of canonically answered questions has not changed. No new encodings needed.`)
+  // print or return something to let calling script know nothing changed, not to commit?
+  process.exit(0)
```
Aprillion, line 33: 0 is a success code, let's try to exit 1 to indicate error so github action would not continue:
```suggestion
  process.exit(1)
```

Aprillion (approved):

questions-encodings.mjs
Aprillion: aah, that worked for skipping commit & push, but failing this step made whole action fail too:

Screenshot 2022-07-01 at 20 18 35

I tried to experiment, but didn't find anything better right now => #41 to improve

Issue #16, opened 2022-06-30
Feedback link on the bottom


plexish:

plexish:
https://docs.google.com/forms/d/1S5JEjhRE8H8MecJuE-X066akb93HUCNFQIVxBhxf2GA/edit?ts=62be9caf

Issue #18, opened 2022-06-30
Switch to colorful icons for Share/Link/Get involved


plexish:
https://www.figma.com/proto/gm06d1l5stMlnKxeElWBEc/Stampy-AI-Alignment-FAQ?node-id=116%3A415&scaling=min-zoom&page-id=116%3A414&starting-point-node-id=116%3A415

PR #44, opened 2022-07-02
Add colored banner, icons + other styling

ccstan99:
I added the colors to the background banner and top icons. Reordered/renamed some links and added the feedback link at the bottom. I haven't really tested on dark mode or smaller sized devices, so probably broke something there. I'm not very familiar with clamp and relative viewport sizes, so I'll probably need your help to clean those up. Here's a screenshot of what it looks like for me. If yours is radically different, please let me know if I goofed up something major.
Screen Shot 2022-07-01 at 11 34 13 PM

I'm still tweaking font sizes, styles, and colors. I've also temporarily hardcoded the intro. I'd like the link to AGI on its own line when it's wide enough but wrap when it's narrow, i.e. I don't want:

I can answer your questions about artificial general intelligence
safety

I can answer your questions
about
artificial general intelligence
safety

Finally, if the overall GitHub action is going to fail, I think it's better to just generate the encodings and let git commit decide if the 2 files are the same. It only takes a few extra seconds to generate the encodings. Just comparing number of questions risks us not catching changes if an existing question is reworded or the number of questions added == removed.

Resolves issue #16 #18 #30


ccstan99:
Oh, another thing... hovering over the text link for share/code/get involved doesn't swap to the rollover for the image. I'm a bit tired to dig into it now. If it's a quick fix for you, I'd appreciate the help. Otherwise, I'll continue working on it tomorrow. Thanks!

Aprillion (approved):
LGTM on a first look, but going to catch a train, then the pride parade in London, then the plane, then some sleep, so might I would get a longer look on the way or later - feel free to merge if you test it more before I get to it :D

Aprillion:
Firefox, Chrome, and other browser dev tools can emulate of [mobile viewports](https://developer.chrome.com/docs/devtools/device-mode/) and [dark theme](https://developer.chrome.com/blog/new-in-devtools-102/#emulation).

Until we find better colors for dark mode, the inverted scheme looks good-enough to me. I will investigate how to show the colorful icons only on hover.

Screenshot 2022-07-02 at 10 51 58

Aprillion:
Updated to only show colors on mouse hover, using the existing svg files:

Screenshot 2022-07-02 at 12 03 42

ccstan99:
Oh sorry, I didn't make myself clear. I had the image hover generally working the way I liked -- normally dull color (not completely black & white) then full color on hover. The inverted colors on dark actually look ok to me. What didn't work was hovering over the text/link below would only underline but didn't change the associated image.

Bigger issue now is that the UI doesn't work at all on iPhone. Tried Chrome and Safari. I'll create a new issue with screenshots and stack trace. I'm guessing the emulator didn't flag that? I'm not sure how to begin addressing that besides slowly backing through our updates this last week to see when it broke. If I had to guess, maybe the tfWorker thread stuff?

ccstan99:
Ah, I see you already fixed the colored image when hovering over text link. Thanks! Safe travels home!!

ccstan99:
BTW, I don't quite understand how all that's working yet but super cool to generate the icons automatically! I'll have to test it out in a bit since I'll be coloring the pencil and other icons. Very minor issue, out of curiosity, would it be easy to do a fade in transition from the black & white icon to color over a period of 1s?

Aprillion:
HTML files in a browser (as well as JSX syntax in React) natively support inline SVG elements, so the [svgr](https://react-svgr.com/) tool just takes the svg file content, runs some optimizer on it and wraps it into an exported React component - we could copy&paste the raw `<svg>...</svg>` directly to our own components and it would work the same, the benefit of the tool is that we can export new version of svg files from Figma or other design tools without manually updating the copied `<svg>` tags, we can just regenerate the components.

Inline svg supports all kinds of cools stuff, CSS styling includes all the transitions that are possible in HTML, as well as more animation possibilities that are SVG-specific (those can be controlled from JS, e.g. to trigger the stampy logo animations in the future, we can define all those animations inside SVG - https://css-tricks.com/guide-svg-animations-smil/ looks like a good intro). Not sure how much is supported by design tools, but if we finalize the shape and colors of the icons/logo, we can decide to update SVG source code manually to craft those animations by hand (we would have to stop generating the files from exported SVG though, any future updates would have to be manual I guess).

ccstan99:
Thanks for the info! I'll look into the SVG animations some more. BTW, should I delete all the icons we're not using?

Issue #19, opened 2022-06-30
"Helpful" button


plexish:
Increment a counter in coda, don't let people submit twice without refreshing, but accept that people can mark it multiple times in different sessions as we don't have session tracking?

ccstan99:
Like button title/tooltip should say "like question" 
Icon should be black & white by default and full cover on hover.
https://github.com/StampyAI/stampy-ui/blob/master/app/assets/icons/thumbs-up.svg
See pen icon for implementation reference.
<img width="328" alt="Screen Shot 2022-07-03 at 2 15 08 AM" src="https://user-images.githubusercontent.com/38997995/178089207-e554bb03-3f5d-45b3-ade4-b60b6ec03954.png">

plexish:
not "Like answer"?

ccstan99:
@plexish You're right! I fixed the instructions above.

plexish:
Decided to go with the name of this button being "Helpful"

PR #48, opened 2022-07-03
colored action icons, fix small screen layout

ccstan99:
A bunch of styling stuff here: Colorized action icons & generated React components. Kept all icons black & white by default but full color (not inverted even on dark mode) on hover. Left all the icons for you want to implement but you can remove them until then. For some reason I need to reload the browser a couple times before the action icon outlines appear at the bottom of each questions. Not sure what's wrong there.

Screen Shot 2022-07-03 at 2 14 54 AM
Screen Shot 2022-07-03 at 2 15 08 AM

Also fixed the header layout to adjust for resizing viewports especially mobiles. When small, icons are spaced horizontally under logo & intro. Replaced text character with chevron for expanding questions. 

Screen Shot 2022-07-03 at 2 29 16 AM

Started styling the search a bit (removed the empty/no results messages) but haven't put in the magnifying glass yet. When the user starts typing before search results are returned there's some funky artifact that still needs cleaning.


app/components/question.tsx
```diff
@@ -3,7 +3,15 @@ import AutoHeight from 'react-auto-height'
...
-import iconPen from '~/assets/icons/pen.svg'
+import {
+  ThumbsUp,
+  Pen,
+  CommentQuestion,
+  Flag,
+  EyeSlash,
+  Link,
+  Ellipsis,
+} from '~/components/icons-generated'
```
Aprillion, line 14: 👍

app/components/question.tsx
```diff
@@ -80,12 +88,18 @@ export default function Question({
                 ref={answerRef}
               />
               <div className="actions">
+                <ThumbsUp />
```
Aprillion, line 91: ah, we shouldn't show the thumbs up button until #19 is implemented to store the upvote, and that one is probably dependent on adding that feature to stampy's Flask server without directly calling the pageforms autoedit MediaWiki API (which has too much power to open up for anonymous use) => it was moved to "gamma" milestone (~August?)

@plexish do you know what other action icons are ready besides the Edit button?

app/components/search.tsx
```diff
@@ -68,18 +69,18 @@ export default function Search({onSelect}: Props) {
...
+        placeholder="Some starting questions below. Type your questions here..."
         onChange={(e) => handleChange(e.currentTarget.value, questions)}
         onFocus={() => setShowResults(true)}
         onBlur={() => setShowResults(false)} // TODO: figure out accessibility of not blurring on keyboard navigation
       />
       <div className={`dropdown ${showResults ? '' : 'hidden'}`}>
         {results.length > 0
           ? results.map(({title, score}) => <ResultItem {...{title, score, model, onSelect}} />)
-          : inputRef.current?.value && <div className="empty">(no results)</div>}
+          : inputRef.current?.value}
```
Aprillion, line 83: Please remove the whole conditional expression if we don't want to display `(no results)` at all:
```diff
-        {results.length > 0
-          ? results.map(({title, score}) => <ResultItem {...{title, score, model, onSelect}} />)
-          : inputRef.current?.value && <div className="empty">(no results)</div>}
+        results.map(({title, score}) => <ResultItem {...{title, score, model, onSelect}} />)
```

Aprillion (changes requested):

app/components/question.tsx
ccstan99: Should I remove everything except for the Edit button for now?

app/root.css
```diff
@@ -164,22 +152,24 @@ h3 a:visited {
...
-  stroke: #333 !important;
-  stroke-width: 10 !important;
-}
 
-.icon-link {
-  transition: all 1s ease;
+.actions svg:not(:hover),
+.icon-link-group .icon-link:not(:hover) svg {
+  filter: var(--light, invert(0)) var(--dark, invert(1));
 }
```
ccstan99, line 173: Does this look ok to you? Any ideas why I have to reload the page a few times before the outlined action icons show up. I shouldn't merge this into the main branch until that gets sorted out.

app/root.css
```diff
@@ -257,24 +247,23 @@ article > h2 > .transparent-button {
...
-  content: '›';
-  margin-top: -8px;
-  font-size: 2em;
-  line-height: 0;
-  font-weight: normal;
+  content: url('/assets/chevron-right.svg');
+  width: 12px;
+  height: 12px;
+  filter: var(--light, invert(0)) var(--dark, invert(1));
   transition: transform 400ms;
```
ccstan99, line 254: The chevron size doesn't stay 12x12. It shrinks when the screen is small and questions wrap onto the 2nd line. How would you recommend fixing? Should I style it in JSX instead of using CSS ::after?
Screen Shot 2022-07-03 at 1 36 28 PM

app/components/search.tsx
ccstan99: Should I remove all the references to inputRef also?

ccstan99:
Tweaked the layout a bit more. Couldn't style the intro when fetched from wiki so reverted back to hard-coding the intro. For related questions, increased left indent, lengthened animation duration, shortened distance of movement, and fade in.

Made wording clearer that search bar is for searching for more questions until better styled and load more questions implemented.

Removed icons from page for actions yet implemented. I haven't been able to figure out how to get the outline on the question actions reliably (weird loading issue) and would like hover on dark mode to be regular full color invert(0). So this should be a more usable version to merge into main branch for production.

ccstan99:
Just checked related questions animation on the iPhone. There's a flicker showing the final state before the sliding animation. Yuck! So unless you know a quick fix for this and colored icons stuff, I'll create a couple of issues for those when merged, so we won't forget later.

app/components/question.tsx
Aprillion: yeah, I would remove everything except Edit for now

app/components/search.tsx
Aprillion: yes please, good catch

app/root.css
Aprillion: if the parent `h2` element is still `display: flex`, we need to add `flex: none;` to the chevron to stop it from shrinking - https://developer.mozilla.org/en-US/docs/Web/CSS/flex#values

ccstan99: Ah, that was easy! I knew you'd know off the top of your head without me digging around.

ccstan99:
Got the chevron size fixed and removed the unused inputRef. Please let me know if you're ok with merging this into the main branch so the changes can be deployed for the Less Wrong post. Thanks!

app/root.css
Aprillion: Hm, not sure, I will investigate the action icons hover.

Aprillion: Looks like I added the `link-icon` class to svg element itself by accident, should only be on the parent `a` and `button` elements, and then in CSS detect the hover on that parent: `.link-icon:hover svg`. Now updated to show only the outline without hover (and outline is inverted in dark mode). I also removed the inverted colors on hover in dark mode:

Screenshot 2022-07-05 at 10 39 38

Aprillion: actually, I was still able to reproduce the problem that I don't see the outline after I reload the page without any question open, then open a question :/

hmm, I need to investigate more

Aprillion (changes requested):
Note to self: I still need to investigate the missing outline in action icons after lunch...

app/root.css
Aprillion: fixed in https://github.com/StampyAI/stampy-ui/pull/48/commits/efbb90b32357d9324ce4c56a91baefe6258dc332

Aprillion (approved):
I fixed action icons, removed unused theming feature and cleaned up the styles and unused code. Deployed to ui.stampy.ai already.

I also added a transition for the icons hover and found out how to keep the lines with `#333` original color inverted even on hover, while preserving the other colors non-inverted.
Screenshot 2022-07-05 at 14 52 31



Let's merge this for now, feel free to improve styles and stuff in a future PR.

ccstan99:
Awesome, thank you so much!!

There was a comment//question earlier up in this thread about which action icon feature can be implemented. I think "Hide" (eye-slash) is removing the question from the current list on screen. That shouldn't require a wiki call since it's all within the UI. Ideally/eventually, users should be able to drag & drop to reorder questions and drag it off the screen to deleted. :)

PR #56, opened 2022-07-06
make search selection idempotent to fix #53

### LINKED ISSUE #53: Remove selected Q from search dropdown
After a user has selected a question from the search dropdown menu, it should be removed from the list so they can't try selecting it again, which currently causes the system to get stuck.
Screen Shot 2022-07-05 at 11 24 02 AM

Aprillion:
* clicking on the same search results multiple times will move it to the top position only once

### DIFF
--- a/app/root.css
+++ b/app/root.css
@@ -353,10 +353,6 @@ footer > *:not(:last-child) {
   box-shadow: 0 0 8px rgba(0, 0, 0, 0.3);
 }
 
-.result-item {
-  line-height: 1;
-}
-
 .dropdown {
   color: black;
   background-color: #f6f6f6;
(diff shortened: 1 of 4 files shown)


app/root.css
```diff
@@ -378,10 +375,16 @@ footer > *:not(:last-child) {
...
 
-.dropdown button:hover {
+.result-item:hover {
   background-color: #ddd;
 }
 
+.result-item.already-open {
+  color: gray;
+  cursor: default;
+  background-color: transparent;
```
Aprillion, line 385: @ccstan99 I added a way to style search results for questions that are already open - this way:
* it won't go away during multiselect (Shift+click)
* we don't have to return more than 5 results from the worker for filtering
* clicking on grayed search result still brings it to top of page as an escape hatch for people who would gets lost and used the search feature to find what is already "somewhere" on the page

Feel free to customize the styles and/or create follow-up GH issues for better functionality (e.g. different color for all previously visited questions in this browser, not just the currently open ones).

already open

PR #62, opened 2022-07-09
Filter out unanswered questions

Aprillion:
No more related questions without canonical answer + some animation improvements (sorry, the 2 second long sliding was driving me crazy, I had to make it faster, but it starts with a delay after opening the question) + I added `eslint` to the project when I noticed some missing dependencies in `useEffect`

without unanswered


ccstan99 (approved):
Looks great! Just a small tweak to shorten the distance of sliding questions.

Issue #42, opened 2022-07-01
Deployment to ui.stampy.ai


Aprillion:
Add github action for deployment to cloudflare worker - see README for instructions to update wrangler.toml with following public keys before running `npm run deploy`:
* account: 841738ad455064a5846675cc41124c85
* KV: 3efee496497048eca81d3f1b036d2631

..see e.g. https://github.com/marketplace/actions/deploy-to-cloudflare-workers-with-wrangler

Contact @Aprillion to generate the `CF_API_TOKEN` from @plexish's CF account.

ccstan99:
I tried to use the wrangler-action workflow from Cloudflare but got an error:
✘ [ERROR]The entry-point file at "build/index.js" was not found.
<img width="820" alt="Screen Shot 2022-07-10 at 6 56 04 PM" src="https://user-images.githubusercontent.com/38997995/178204706-cefd8c82-28dc-4483-ac2d-39d740828916.png">

So, I just used our regular `npm install` then `npm run deploy`

Instead of hardcoding those values above, I used GitHub's repos secret keys: CLOUDFLARE_API_TOKEN, CLOUDFLARE_ACCT_ID, STAMPY_KV_ID

Aprillion:
Cool! Did you set up the github secrets with plex already? If not, I can have a look later after work.

ccstan99:
We can use the same action/workflow and add a trigger on PR merge/close. So we might as well generate the encodings at the same time. It takes less than a minute and ensures everything is always up-to-date.

Right now the triggers are manual (workflow_dispatch) and schedule. I didn't put it the PR trigger to avoid throwing in too much all at once. It looks the deploy manual deploy worked.
https://github.com/StampyAI/stampy-ui/runs/7286918471

I'll try to test out the PR merge/close from my repo later. So we should wait until that's working before closing out this issue.

Aprillion:
Merge of PR #71 seems to triggered the action from merging a PR successfully - https://github.com/StampyAI/stampy-ui/pull/71/checks

But it failed because we introduced a rule against pushing code directly to master => I will relax the rule for now and let's see next PR.

ccstan99:
Do you remember that there was originally a force push parameter in the GitHub Action? I think it was to get around the current error we're seeing in the log, based on https://github.com/stefanzweifel/git-auto-commit-action/issues/71. Looks like you can "Restrict who can push to matching branches" while also "Allow force pushes for all users with push access."

[...]

PR #70, opened 2022-07-11
deploy to Cloudflare Worker

ccstan99:
Addresses #42. Since we'd need to deploy AFTER generating encodings and we probably should check for encoding changes BEFORE deploying anyway, I've combined both processes into a single job in the same workflow. I renamed the yml file since it generates encodings also deploys now. Tweaked the wrangler.toml.template to make search/replace easier and updated README.md to reflect changes. Tested deployment to stampy-ui.ccstan99.workers.dev. Sample output from successful run https://github.com/ccstan99/stampy-ui/runs/7276840669

**When merging this with the main branch, 3 secret keys must be created in the GitHub repo:
CLOUDFLARE_API_TOKEN, CLOUDFLARE_ACCT_ID, STAMPY_KV_ID.** I don't have permission to do that.

Screen Shot 2022-07-10 at 6 46 35 PM

This is currently scheduled to run daily at midnight UTC. Checking my logs, it didn't seem to run very reliably but seems to be a [known issue](https://upptime.js.org/blog/2021/01/22/github-actions-schedule-not-working/). We can (probably should) also have this run whenever a pull request is merged with main also. Here's the [syntax](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-when-a-pull-request-merges) if/when you decide that's what you'd like to do.

While you're updating the repo profile, I snuck in a new image for the social media preview if you'd like to update it. https://github.com/ccstan99/stampy-ui/blob/master/app/assets/stampy-ui-preview.png


Aprillion (approved):
I created the CF API Token and added the 3 GH repo secrets, let's try it

PR #71, opened 2022-07-11
trigger on pr merge/close

ccstan99:
Added trigger on PR close. Tested on my repo's main branch. Screenshot of run by 3 triggers: manual, scheduled, and PR merge/close. 
Screen Shot 2022-07-11 at 11 44 31 AM

It doesn't seem to be called when a PR is closed without merging. I included in a couple checks to deploy only if encodings changed or a merge occurred. Might be harmless to remove and deploy even when no change... ie when scheduled. Anyway, left it there for you to decide how/which you want to use.

### DIFF
--- a/.github/workflows/gen-encodings-deploy.yml
+++ b/.github/workflows/gen-encodings-deploy.yml
@@ -3,6 +3,9 @@ on:
   workflow_dispatch:
   schedule:
     - cron: "* 0 * * *"
+  pull_request:
+    types:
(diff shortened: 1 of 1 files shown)


ccstan99:
Forgot to include that this addresses #42. I know this won't auto-close the issue, which is fine. We need to check that things work after the merge anyway.

Aprillion (approved):

Issue #29, opened 2022-06-30
Do caching in the background


plexish:
See TODO in https://github.com/StampyAI/stampy-ui/blob/master/app/stampy.ts#L148-L149

PR #84, opened 2022-08-05
Caching in background

Aprillion:
Fixing #29 and related bugs I noticed while changing the cache.


ccstan99 (approved):
Works really nicely! I didn't check if it refreshed after 10 minutes but the code & KV values look good to me.

Slightly side issue, should there be an error boundary somewhere for handling unforeseen hiccups?

Aprillion:
@ccstan99 thanks! Good idea about the error boundary => #87

Issue #92, opened 2022-08-16
Debounce search


Aprillion:
Tensorflow search can take a few 100 milliseconds to complete. When typing fast into the search box, we shouldn't search for each intermediate partial string, but we should wait while typing and only search after e.g. 300ms.

We can use https://lodash.com/docs/4.17.15#debounce

Issue #32, opened 2022-06-30
Make plaintext search run until TF is loaded


plexish:
Actual:
* after fixing #33, we show `baselineSearchResults` whenever `searchResults` list form TF is empty and will switch to TF results immediatelly when available

Expected:
* show `baselineSearchResults` only while TF is not loaded, then rely fully on TF
* do not flicker the results for the current search term, switch only for future queries after the initializaiton

PR #93, opened 2022-08-16
small fixes

### LINKED ISSUE #91: Some websites parse out the trailing underscore
e.g. If you post a raw link to an answer to github like
https://ui.stampy.ai/?state=8157_

The _ gets stripped, so the link does not work. Can we detect when the last pageid is missing a control character and re-add it? Or just assume open if the control character is missing?

Aprillion:
fix #91, #92, #32, and some chevron alignment that caught my eye:
Screenshot 2022-08-16 at 13 41 08


ccstan99 (approved):
Looks good to me. `debounce` seems very handy! Sounds like `throttle` should come in handy for the infinite scroll later?

I'm still not very good with TypeScript, so just curious, what's the motivation behind: `const empty: [] = []`?

Aprillion:
> what's the motivation behind: `const empty: [] = []`?

aah, that is a left-over from a different angle to use `setBaselineSearchResults(empty)` with a stable array reference to prevent unnecessary re-renders after TF search finished loading .. but in the end I had to use `tfFinishedLoadingRef` for mutable communication between the event handler and the response from worker, so no good reason to use `empty` instead of literal `[]` in current code (but not harmful either, it will just behave the same)

the TS reason to use an explicit type for empty array is that `const empty = []` has an implied type `any[]` and our eslint rules complained about implicit `any`

ccstan99:
Thanks for the explanation! Yeah, I didn't think there was anything harmful, just wondering if there was some convention or best practice that I should be aware of.

PR #95, opened 2022-08-20
Add search icon and link to question

Aprillion:
Both icons start as outlines and show the fill on mouse hover.

preview with new icons


ccstan99:
The changes mostly look good but I tinkered with the table styling a bit more to make the rows top aligned, outlined cells, header text small caps, add a hint of background coloring at https://stampy-ui.ccstan99.workers.dev/. I can add commit to this branch if it looks ok to you & @plexish.

Couple of questions:
- Is it important to have the rows underlined when hovering, since the user can't really interact with the content? I removed it but if you want the feature, I'd prefer highlighting the entire background rather than just underline.
- For some reason, I thought the link to this question icon would copy to clipboard like the share link but with only that question. As implemented, it goes to the page with just that question. Of course, the user can then copy the URL and essentially accomplish the same. If I misunderstood, no worries.
- Finally, the magnifying glass color-change behavior feels a bit odd. It changes when cursor is over/in the input box but reverts to b&w when the cursor is over the icon (presumably because it's not clickable). I'm tempted to always leave it black & white. Thoughts? Also, when you type something in the input box, an 'x' appears behind the magnifying glass. Anyway, we can discuss as the next office hours or I can write these up as an issue.

And sorry you had to fish out the old icons, I must have accidentally removed them from the repo when I was tidying up before.
Screen Shot 2022-08-22 at 5 59 59 PM
Screen Shot 2022-08-22 at 6 17 49 PM

Aprillion:
oh, you copied the discussion on discord too - I answered there.

ccstan99 (approved):
OK, I'll merge this now and create a new PR & issue(s) based on our discussion on Discord.

Issue #98, opened 2022-08-23
Search icon styling


ccstan99:
Search icon should remain black & white. Also style with :`:-webkit-search-cancel-button` pseudoelement for browsers showing the 'X'

Issue #97, opened 2022-08-23
Link button should copy to clipboard


ccstan99:
Clicking on "Link" icon button should copy the link to clipboard instead of going directly to the page.

plexish:
Maybe give a little visual indication that this has happened too? Saying "Link copied to clipboard"?

Aprillion:
Left click will now (in the PR) copy the link, while it's still possible to use context menu or middle mouse button to open the link on new tab. Same confirmation as for the `Share link` header icon, I changed the message to "Link copied to clipboard" for both.

PR #100, opened 2022-08-25
Fix search icon and copy link

Aprillion:
Fixing #98 and #97:
new behaviour


ccstan99 (approved):
Looks great, especially neat that the link using the context menu works too!

PR #103, opened 2022-09-08
Tweak title, background, icons for noLogo

ccstan99:
When noLogo=true, the window title is set to "AI Safety FAQ", no colorful background shown, and place top icons in a row to minimize vertical space. As always, LMK if there's a better way to do this.
https://stampy-ui.ccstan99.workers.dev/?noLogo=true
Screen Shot 2022-09-08 at 1 41 30 PM

### DIFF
--- a/app/hooks/useQuestionStateInUrl.ts
+++ b/app/hooks/useQuestionStateInUrl.ts
@@ -1,6 +1,6 @@
 import {useState, useRef, useEffect, useMemo, useCallback} from 'react'
 import type {MouseEvent} from 'react'
-import {useSearchParams, useTransition} from '@remix-run/react'
+import {useSearchParams, useTransition, useHref, useResolvedPath} from '@remix-run/react'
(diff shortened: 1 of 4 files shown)


Aprillion (dismissed):
this approach should work too, but I think a conditional class might be more appropriate than inline styles here, something like:

```css
header.withLogo  {
  background-image: url('/assets/coded-banner.svg');
  background-repeat: no-repeat;
  background-attachment: fixed;
  background-position: top center;
  background-size: 800px;
}
```

```jsx
<header className={noLogo ? 'noLogo' : 'withLogo'}>
```

Aprillion:
Updated with the above suggestion + detection of domains as discussed on discord, only `ui.stampy.ai` has the logo by default and other domains are without logo by default, we can force which version is used by using `noLogo` or `withLogo` in the path.

Aprillion (approved):

PR #105, opened 2022-10-18
fix GitHub Action to auto generate embeddings

ccstan99:
@plexish @Aprillion I dug around a bit and I *think* the GitHub Actions for generating the embeddings was failing because the script doesn't have permission to push the generated embeddings to the protected main branch. According to (https://github.com/marketplace/actions/push-to-status-check-protected-branches), we might be able to work around this using PAT (personal access tokens). Since I don't have permission to do this, could one of you please:

(1) Create a PAT which grants the following scope/permission with admin write:repo_hook to stampy-ui (https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token)

(2) Create a repository secret named "PAT_GEN_EMBEDDINGS" with the value of the PAT from the previous step (https://docs.github.com/en/codespaces/managing-codespaces-for-your-organization/managing-encrypted-secrets-for-your-repository-and-organization-for-github-codespaces)

In this PR, I've updated the GitHub Action to use that new PAT stored under PAT_GEN_EMBEDDINGS. Hopefully, this will work. Otherwise, I found an alternate solution but it felt a bit doesn't look very "clean." Anyway, if all goes well, we set this up to run automatically.

### DIFF
--- a/.github/workflows/gen-encodings-deploy.yml
+++ b/.github/workflows/gen-encodings-deploy.yml
@@ -34,7 +34,7 @@ jobs:
         if: steps.check.outputs.changed == 'true'
         uses: ad-m/github-push-action@master
         with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+          github_token: ${{ secrets.PAT_GEN_EMBEDDINGS }}
           force: true
       - name: Restore stampy-ui repository code
         if: github.event.pull_request.merged == 'true' || steps.check.outputs.changed == 'true'
--- a/questions-encodings.mjs
+++ b/questions-encodings.mjs
@@ -1,6 +1,7 @@
 // server side script called by GitHub Actions to periodically generate encodings stampy.ai wiki questions
 
 // npm install node-fetch @tensorflow/tfjs-node @tensorflow-models/universal-sentence-encoder
+// node questions-encodings.mjs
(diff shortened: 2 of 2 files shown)


Aprillion (approved):
token added to PAT_GEN_EMBEDDINGS repo secret

I also updated your role to admin, so you will be able to make this kind of changes yourself in the future

Aprillion:
https://github.com/StampyAI/stampy-ui/actions/runs/3271257190/jobs/5380822567 still failed with error `At least 1 approving review is required by reviewers with write access.` when using `admin:repo_hook` permission :(

When I deleted my personal token and [re-run the action](https://github.com/StampyAI/stampy-ui/actions/runs/3271301498/jobs/5380923227), the error message was `Invalid status code '401'` => the token itself worked, we would have to disable the protected branch approving reviews check to make it work (and if I remember correctly, GITHUB_TOKEN would be enough, no need for my personal token if we decide to do that)

ccstan99:
Bummers! I'm hoping we won’t need to disable approving reviewers. That seems like an unnecessary security risk. Just to make sure, the PAT had "write" access to admin:repo_hook, correct? The alternative hack was to use a user with admin access. I can test it out if you’ve already granted me the admin role. I’m going to bed now and have other obligations in the morning but should be able to dig further into this after that. Thanks for taking care of things up to this point!

PR #107, opened 2022-10-19
Fixes to auto-gen encodings and others

ccstan99:
I created the PAT_GEN_EMBEDDINGS in the repo secrets. Didn't see it there before. Also added a couple lines to `persist-credentials` so we can force push without PR with reviewer approval. I tested this on my own by also requiring reviewer approval. Got it to fail before then worked after fix:
Screen Shot 2022-10-18 at 8 33 26 PM
Screen Shot 2022-10-18 at 9 30 49 PM

Removed the note about generating React components for icons in the README. It seems to do that automatically, probably more often the necessary. Maybe can be moved elsewhere if needed? Just want to keep the key info in README. That didn't seem like critical information for onboarding new devs & getting them functional speed quickly.

Since there's been quite a few updates, the latest encodings are here ... just in case the GitHub Action STILL doesn't work.


questions-encodings.mjs
```diff
@@ -7,6 +7,22 @@ import * as use from '@tensorflow-models/universal-sentence-encoder'
...
+  titles: 'Canonically_answered_questions|Canonical_answers|Canonical_questions',
+  format: 'json',
+}
+const options = {
+  method: 'POST',
+  body: JSON.stringify(params),
+  headers: {'Content-Type': 'application/json'},
+}
+await fetch(urlPurge, options)
+
```
ccstan99, line 25: According to @plexish's notes and documentation, I made a POST call to purge the wiki cache. It doesn't give an error but it's not clear that it really works. The JSON response from fetch seems incomplete.

.github/workflows/gen-encodings-deploy.yml
```diff
@@ -11,6 +11,9 @@ jobs:
     steps:
       - name: Clone stampy-ui repository code
         uses: actions/checkout@v3
+        with:
+          persist-credentials: false # otherwise, the token used is the GITHUB_TOKEN, instead of your personal access token.
```
Aprillion, line 15: aah, it recycled the credentials from this step 🤦 thanks for having the patience to investigate 🙇

Aprillion (approved):

Issue #106, opened 2022-10-18
Investigate caching issue


Aprillion:
We discovered that cache refresh doesn't work - e.g. for https://ui.stampy.ai/?state=8241_ (=> KV key 8241 in https://dash.cloudflare.com but CF doesn't show metadata of KV records, so I couldn't quickly check whether here is a problem in expiration date format for that particular record, for all records, or something else)

I cleared the KV store records as a quick fix, but need to follow up on the investigation when I (or someone) have more time.

Ideas that could help to prevent hard-to-debug caching issues in the future:
* move timing info from metadata to KV value itself (to make it inspectable in dashboard)
* add Remix API route to clear the cache (and maybe redirect to homepage in order to prime the cache at this point for fast initial load for the next user)
* replace the stale version we display in UI when a new version of page is loaded on the server (do not require manual reload)
* try to reproduce with some Test question:
   * load page
   * edit page on wiki
   * wait for the stale time (ideally decrease on localhost from 10 minutes to 10 seconds for this investigation)
   * reload page
   * check logs/debugger - did we fetch a new version in the worker and update the cache?
  
...and if is the problem reproducible even on localhost, let's fix it 🤞

Aprillion:
I added https://ui.stampy.ai/questions/cache endpoint to see all KV store metadata and all timestamps are from October 18 => refreshing data in the background does not work, seems like real CF workers kill the background requests, and it only ever worked on localhost (using `wrangler` and `miniflare` to simulate workers as recommended by CF, but there are differences it seems).

So... if I want to refresh in the background, I have to make the request from browser (and I guess I should replace the content once the new content is loaded).

Workaround until fixed = manually load https://ui.stampy.ai/?reload (but it only works for homepage at the moment unfortunately)

Aprillion:
I haven't heard @plexish run around and scream about any new caching issues, not even after https://www.lesswrong.com/posts/fSMrwJnqRb5NrMYFx/all-agi-safety-questions-welcome-especially-basic-ones-2 ..

Checking https://ui.stampy.ai/questions/cache, there were still some left-over static assets I investigated in #110, #111, #112, but the redirect is removed and I didn't experience troubles since, so cleared the cache and will optimistically hope that I won't see any static assets on that page in the future (and if yes, I will open a new issue to follow up on that investigation)

Aprillion:
[...]

Issue #94, opened 2022-08-19
Use `defer` in remix loader instead of fetching from `useEffect`


Aprillion:
We load low priority data by using `fetch('/questions/allCanonicallyAnswered')` in a `useEffect` in https://github.com/StampyAI/stampy-ui/blob/master/app/hooks/useQuestionStateInUrl.ts#L40

I got some tips from Kent in https://youtu.be/0VgyPIRHZl8?t=2457 (starting 41:00)

Since CF Workers support streaming responses, we can send low priority data from a loader without blocking the initial screen:
https://github.com/kentcdodds/advanced-remix/blob/main/final/09-defer/app/routes/__app/sales/customers/%24customerId.tsx

Let's update our code to load the data as soon as possible without blocking initial SSR render, without waiting for client-side `useEffect` - `defer` looks like the best solution (we also have to update to React 18, hopefully all hydration issues will be resolved).

plexish:
Is this done @Aprillion?

Aprillion:
Not done, but I'm actually happy with fetching from useEffect, not missing any features provided by `defer`, so let's close this one and if we need to use it for some feature/bugfix we can use it later.

PR #109, opened 2022-10-29
Cache

Aprillion:
move background fetching for cache reload to browser

* CF workers kill hanging fetch requests, so previous version didn't run in production, only during local development
* fixing #106

Client code is getting more complex, but hopefully we can move the caching logic back to "server" in #94 if we start using HTTP streams...


ccstan99:
I tested the changes for caching locally and everything looked great. However, when I deployed it to my test server, I just got header & footer with no content (empty list of questions), presumably because the already stored cache data format is different from what is now expected. After using "?reload" in the browser, I got: `TypeError: Cannot read properties of undefined (reading 'pageid')` Only the data for the key `initialQuestions` was updated but nothing else. After I cleared the entire KV cache, things worked fine. Anyway, I'm hesitant to approve & merge myself (which will auto-deploy) in case it'll cause an issue transitioning to the production websites since I don't have access to fix those. I'd be happy to proceed if/when you're prepared to be available to manage things on the CF side...

Aprillion:
hmm thanks for testing - I tried to make the transition work, but only tested it once in the middle of my changes and then everything was migrated and I made more changes 🤦 

I will have to deploy the old version to my test server and try the "migration" again

ccstan99:
Ideally, if there's some way to purge the STAMP_KV cache whenever we deploy a new version to CF, that might be a safer way to start with a "clean slate."

Aprillion:
Good idea, there will be a button `Clean cache` on https://ui.stampy.ai/questions/cache (preview: https://stampy-ui.aprillion.workers.dev/questions/cache) + I added a step to github action to send the POST request to clean cache, let's see if it works.

ccstan99 (approved):
Thanks so much for taking care of all of this! Wishing us luck for a smooth deployment.

ccstan99:
GH Action failed and https://ui.stampy.ai/questions/cache doesn't seem to work. I tested it on my dev CF and it was fine. Fortunately, https://ui.stampy.ai/ and https://alignment.wiki/ still work, for now.
Screen Shot 2022-10-30 at 12 24 39 PM
Screen Shot 2022-10-30 at 12 28 51 PM

Issue #37, opened 2022-06-30
"None of these: Request an answer to my exact question above"


plexish:
Add to questions table in coda

Aprillion:
#114 is a temporary solution with a link to wiki, @plexish wanted to have anonymous access to request new question from UI for the full solution of this issue.

plexish:
Or via github secrets allowing the UI to log into the wiki directly to add the question, that seems fine too if we want to touch less code.

plexish:
Re-use Stampy account from bot, probably use github secrets.

plexish:
@mruwnik did this in #135

PR #114, opened 2022-11-07
add button to request exact question

lcb931023:
The button is an anchor link to `https://stampy.ai/wiki/Special:FormStart?form=Q&page_name=${searchInput}`. This opens the form for adding a new question.

### DIFF
--- a/app/components/search.tsx
+++ b/app/components/search.tsx
@@ -95,29 +95,30 @@ export default function Search({
       </label>
       <AutoHeight>
         <div className={`dropdown ${showResults && results.length > 0 ? '' : 'hidden'}`}>
-          {showResults &&
-            results.map(({title, score}) => (
-              <ResultItem
-                key={title}
-                {...{
-                  title,
-                  score,
-                  model,
-                  onSelect,
-                  isAlreadyOpen: openQuestionTitles.includes(title),
-                }}
-              />
-            ))}
+          <div>
+            {showResults &&
+              results.map(({title, score}) => (
+                <ResultItem
+                  key={title}
+                  {...{
+                    title,
+                    score,
+                    model,
+                    onSelect,
+                    isAlreadyOpen: openQuestionTitles.includes(title),
+                  }}
+                />
+              ))}
+          </div>
           <a
             href={`https://stampy.ai/wiki/Special:FormStart?form=Q&page_name=${searchInput}`}
-            target='_blank' rel="noreferrer"
-            className={`transparent-button result-item`}
-            style={{display: 'flex', alignItems: 'center', textDecoration: 'none'}}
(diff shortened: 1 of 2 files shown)


Aprillion:
what was the issue with this requested, please? I don't remember the exact requirement 😅

lcb931023:
This feature is to allow UI users submit new questions they have in mind, in case Stampy doesn't return a satisfying match.

@plexish mentioned that this is important to work on during weekly meetings in ~October. Though, I was unable to attend the last few ones, so maybe thing's changed :p

The need to allow new question submission was also referenced in https://github.com/StampyAI/stampy-ui/issues/37

plexish:
Yup, this still seems like a key feature.

app/components/search.tsx
```diff
@@ -106,6 +108,17 @@ export default function Search({
...
               />
             ))}
+          <a
+            href={`https://stampy.ai/wiki/Special:FormStart?form=Q&page_name=${searchInput}`}
+            target='_blank' rel="noreferrer"
+            className={`transparent-button result-item`}
+            style={{display: 'flex', alignItems: 'center', textDecoration: 'none'}}
+            key='none-of-the-above'
+            title='Request a new question'
+            onMouseDown={(e) => e.preventDefault()}
```
Aprillion, line 118: what's the purpose of this line, please?

app/components/search.tsx
```diff
@@ -106,6 +108,17 @@ export default function Search({
                 }}
               />
             ))}
+          <a
+            href={`https://stampy.ai/wiki/Special:FormStart?form=Q&page_name=${searchInput}`}
+            target='_blank' rel="noreferrer"
```
Aprillion, line 113: could you please check `npm run lint` and/or enable eslint and prettier in your editor? JSX props should be in double quotes in our code style settings, I believe

app/components/search.tsx
```diff
@@ -106,6 +108,17 @@ export default function Search({
...
+            className={`transparent-button result-item`}
+            style={{display: 'flex', alignItems: 'center', textDecoration: 'none'}}
```
Aprillion, line 115: I would prefer adding a className and corresponding style in the `css` file instead of non-dynamic inline styles

Aprillion (approved):
thanks for pointer to the issue, this looks like a good first step 👍

I have a few minor comments, please fix if it's clear what I'm asking or let me know if not clear and I can fix those at a later time after merging.

Aprillion (approved):
I updated this PR as per my previous comments and I also moved the link below the search results box to distinguish it more from search results:
image

@ccstan99 I did not consult your design for this is a temporary link to the wiki, the future solution for anonymous request of new questions directly from UI might look different..

PR #120, opened 2022-11-30
fix typo in email

ccstan99:
Oops, GH actions failing since email should be string not variable.

[Invalid workflow file: .github/workflows/gen-encodings-deploy.yml#L34](https://github.com/StampyAI/stampy-ui/actions/runs/3583544070/workflow)
The workflow is not valid. .github/workflows/gen-encodings-deploy.yml (Line: 34, Col: 14): Unexpected symbol: 'hello@stampy'. Located at position 1 within expression: hello@stampy.ai


Aprillion (approved):
yeah, I started to see the benefits of JSON simplicity over YAML after debugging these sorts of troubles :D

Issue #35, opened 2022-06-30
Style search results


plexish:
https://www.figma.com/proto/gm06d1l5stMlnKxeElWBEc/Stampy-AI-Alignment-FAQ?node-id=116%3A415&scaling=min-zoom&page-id=116%3A414&starting-point-node-id=116%3A415

ChrisRimmer:
I'm gonna be honest I got a bit excited now that I finally have a viable dev machine running and went ahead implementing this before noticing @ccstan99 is assigned. If you'd like to take a look at the pull request linked above where I had a go at this, it'd be appreciated.

ccstan99:
No worries! I've completely forgotten about this so I'm more than thrilled that you're tackling it. Thanks! I'll take a look later today when I have some time.

mruwnik:
A couple of additional things introduced by https://github.com/StampyAI/stampy-ui/pull/179

1.     Vertical height of dialogue shrinks after switching to thank you message, would be nice to also stay fixed.
2.     There probably should be a more explicit way (x in upper right and/or OK/Close button) to close the dialogue after the question has been requested/suggested. Not too critical since clicking outside closes it anyway.

PR #121, opened 2022-12-04
DNM -Implements #35

ChrisRimmer:
- Adds a layout CSS file containing a load of utility classes using Flex for layout
- Uses some of the above utility classes to twiddle the layout of the "None of these" option in search results
- Moves "None of these" option into search results list so it hides properly when search is inactive
- Move most search result styling onto each search result rather than the dropdown container to match designs in Figma


ChrisRimmer:
Blocked. Needs rebasing to depend on #122 and for that request to be merged

ccstan99 (commented):
The styling changes look great to me. But I'd like to get @Aprillion input just in case he has an opinion on the utility classes

app/stylesheets/layout.css
```diff
@@ -0,0 +1,68 @@
+.row,
+.column {
+	display: flex;
```
Aprillion, line 3: please use https://prettier.io/ in your editor and/or `npm run lint` from command line

app/components/search.tsx
```diff
@@ -109,17 +109,22 @@ export default function Search({
                   }}
                 />
               ))}
+              <a
+                href={`https://stampy.ai/wiki/Special:FormStart?form=Q&page_name=${searchInput}`}
+                target="_blank"
+                rel="noreferrer"
+                className="result-item none-of-the-above row m-spread x-center"
```
Aprillion, line 116: I'm happy if we decided to switch to utility classes, but it would ideally be https://tailwindcss.com/ or some other library that would be known to other developers instead of rolling our own...

and we could even slowly transition from our current system ("semantic" classes I guess) to utility classes by converting only 1 component at a time and not the whole app...

but I am less happy about mixing semantic and utility classes inside the same element withing one `className`

if you promise you will be working on converting the whole file to utility styles, removing the old classes, ifthis is just temporary mixing of classes from 2 CSS files without explicit imports, then feel free to merge

but we could get away with plain CSS without CSS Modules or other more sophisticated solutions only because we kept all CSS in 1 file that is easy to check for name collisions and unused class names,, adding more plain-CSS files that are imported only in root is asking for trouble IMHO 🤔

Aprillion (approved):
I am not convinced about usefulness of `layout.css` right now, but we can discuss during hackathon and/or we can focus on making all new features work first and I can refactor during Christmas, so Approving this PR for merging if you plan to make a lot of changes and use the styles from `layout.css` extensively in next few weeks 👍

app/stylesheets/layout.css
ChrisRimmer: @Aprillion I've just over a week ahead where I expect to work on Stampy quite a lot, and I've found these kinds of utility classes or whatever you want to call them a very good pattern for solving some common problems in CSS / HTML interactions so while I intend to make fairly extensive use of this stylesheet in coming days, even if I didn't I would at least advocate for moving in that direction anyway. Adam Wathan of Tailwind CSS does a better job explaining it than I do: https://adamwathan.me/css-utility-classes-and-separation-of-concerns/

All that being said, it's possible the actually better move is just using TailwindCSS instead.

app/components/search.tsx
ChrisRimmer: Heh, looks like I didn't scroll enough before sending my other reply. In my own experience I've found that one large sheet doing everything becomes tricky in really silly / mundane ways, it just gets hard to scroll around etc. If we'd rather just use Tailwind I'd be ok with that and will get to work on the relevant changes now.

ChrisRimmer:
I am absolutely revealing the extent of my unfamiliarity with Github's review features today, huh

Aprillion (approved):

Aprillion:
looks like this one didn't get traction and would have merge conflicts with `future` branch, so let's close this one and `future` branch for now.

if we decide to use utility css styles, we can open a new PR from top of master later

Issue #77, opened 2022-08-02
Infinite scrolling


plexish:
Scrolling to the bottom would cause a new 10 questions to be loaded, from a pre-defined list [here](https://stampy.ai/wiki/Infinite_scroll_questions).

Aprillion:
@plexish @ccstan99  do we want to remove footer as part of this feature? footers don't make sense with automatic infinite scrolling, it just creates frustration when you want to click a link from a footer and more stuff loads 😭 

..or we can return to #17 if we decide to keep footer

plexish:
Probably remove, yeah. The other option would be to have it slide into view if you scroll up, like the top bar on LessWrong posts?

plexish:
We could have Discord become a floating bottom-right icon invite

PR #122, opened 2022-12-04
Add utility classes in standalone stylesheet

ChrisRimmer:
Adds a layout CSS file containing a load of utility classes using Flex for layout

Provides a set of useful and spec-aligned classes that can be sprinkled on elements to give them desirable flex layout behaviour. Covers most common scenarios around alignment, direction, and sizing.

Will be used by almost any other work I do WRT styling


Aprillion:
this commit is fully included in PR #121, not sure how it would make sense to merge unused styles independently, let's focus on reviewing the changes together in the other PR

ChrisRimmer:
> this commit is fully included in PR #121, not sure how it would make sense to merge unused styles independently, let's focus on reviewing the changes together in the other PR

It's fully included in #121 as that PR and another I'm working on (toward support for #77 - obv not posted yet) both use the same stylesheet, and I didn't want either change not being merged to block the other due to that mutual dependency. If this PR were to be merged, Git would detect that the commit is no longer needed in the other PR and would omit it from #121.

That said, it looks like the other change I was working on is taking a bit more work than I expected so we can probably safely leave this and let any potential PR for #77 be blocked by #121 without much harm.

Issue #123, opened 2022-12-19
Render answers from Google Docs / Coda


Aprillion:
We are moving our "database" of questions and answers from MediaWiki to Coda and GDocs...

First step for the UI is to render a test document, e.g. let's try the [export](https://docs.google.com/feeds/download/documents/export/Export?exportFormat=html&id=10g6U9SL0CBy__wCBTib7_WhB3S3aaFt7Fx1vVgCzg2I) feature
* we decided to read `Rich Text` markdown field from Coda, so we don't need to merge rendering directly from GDoc

Aprillion:
exploring Coda: https://coda.io/developers/apis/v1#tag/Tables, and looks like GET requests work from browser when I am logged in to UI

list of tables in our doc id: https://coda.io/apis/v1/docs/fau7sl2hmG/tables
Initial questions:  https://coda.io/apis/v1/docs/fau7sl2hmG/tables/Initial%20questions/rows?valueFormat=simpleWithArrays&useColumnNames=true

filter by tag: https://coda.io/apis/v1/docs/fau7sl2hmG/tables/Answers/rows?valueFormat=simpleWithArrays&useColumnNames=true&query=Tags:%22superintelligence%22
filter by status: https://coda.io/apis/v1/docs/fau7sl2hmG/tables/Answers/rows?valueFormat=simpleWithArrays&useColumnNames=true&query=Status:%22Live%20on%20site%22

Aprillion:
I added `CODA_TOKEN` to https://github.com/StampyAI/stampy-ui/settings/secrets/actions while writing README instructions for local development with Coda.

Aprillion:
the new `canonically answered canonical questions` list is 99.9% going to be https://coda.io/d/AI-Safety-Info_dfau7sl2hmG/Aiming-for-Exceptional_suCgH#_lu0xQ

ccstan99:
Thanks for keeping these helpful notes for getting "up to speed" on Coda!

It looks like the Manage answers view includes all canonical questions even if they don't have canonical answers yet (see rows 1-24) that have Status like "Not started", "In progress", "In review", etc. So it probably still needs to get filter/queried by Status as "Live on site"?

Aprillion:
I will leave it up to @plexish to explain, he had some half-baked notes on Discord that made me feel I understand the intention, but I am not sure any more 🤔

plexish:
Hi!

[...]

PR #124, opened 2022-12-28
First gdoc

Aprillion:
First step towards migration to Google Docs, keeping on `future` branch until we have actual content to show in this version (and solve million other issues I guess).

Live deploy: https://stampy-ui.aprillion.workers.dev/

FYI @ccstan99 - no idea how to support previous search feature where we only had question titles, so I removed the worker for now and we will either figure out full backend API or some way to generate client-side embeddings that include google doc ids (the worker code is still accessible in git if we choose the latter path)...

FTR @plexish - I don't have plans to include more features in this PR => please comment in #123 for non-code-review discussions.


Aprillion:
Won't be needed after all, we will read markdown version from the `Rich Text` field from Coda.

PR #125, opened 2023-01-05
Fixes #85 - iOS header icons too dark in dark mode

### LINKED ISSUE #85: iPhone dark mode icons not visible
The icons on iPhone in dark mode are not inverted and cannot be seen easily
IMG_4224

ChrisRimmer:
Screen capture of new behaviour can be found in Discord - https://discord.com/channels/677546901339504640/1060546605008298075/1060546670686916659

Looks like Safari on iOS either didn't like the invert filter, or didn't like the selector we used to target the SVG path elements for inverting when dark mode is enabled. I don't have an OSX machine so can't inspect the browser state to say for sure, so that's just a guess.

Regardless of the cause, this commit puts a workaround in place that uses a React media query hook to change the content of the SVG returned by the image components directly, changing their colour depending whether dark mode is enabled or not. This necessitated a new package in the project, namely `react-media-hook`

I also had to slightly change some of the adjecent CSS to remove the now-redundant inverting filter on these path elements, and due to the selector now becoming unreliable with the new changing CSS I elected to just swap the rules that matched on CSS properties for ones that matches a selector class. Feels weird to use a class just to _un_target some behaviour but is what it is.


Aprillion (changes requested):
These are automatically generated files by `npm run generate-icons` which we run during build in the github action deploy, so they would get overriden :(

If we don't want to re-generate these React components from SVG files provided by @ccstan99, but to make these React components the source of truth, please delete the original files, rename the folder from `icons-generated` to `icons` and remove the script `generate-icons` from package.json (both the definition and all calls to that script)

We haven't changed these icons for a long time, so if there are no plans to update the icons soon, I am fine with manual maintenance of these React components.

ChrisRimmer:
Thanks for the heads up, I'll rework that now

ccstan99:
Ewww, updating icons will be undesirably messy but hopefully that won't happen much if at all. I guess it'll be ok to just make the changes as @Aprillion suggested for now. Thanks!

Aprillion (approved):

ChrisRimmer:
> Ewww, updating icons will be undesirably messy but hopefully that won't happen much if at all. I guess it'll be ok to just make the changes as Aprillion suggested for now. Thanks!

Thankfully @ccstan99, I figured a way to keep the generated icons, turns out the SVGer preserves classes that have been added to particular tags inside the SVG so I just added a class in the source file and targeted it in CSS after

ccstan99:
Ah, gotcha! I saw changes to the generated files but I see now that the target is in the CSS. So we'll just need to remember to add those classes if updating the SVG which seems reasonable. Thanks again!

PR #130, opened 2023-01-14
Deploy only from merged PR to master

Aprillion:
* do not deploy when closed by rejecting PR and not from other target branches


Aprillion:
not sure why the GH Actions are exectuted from a branch too, maybe if we merge this PR, it will get cleaned up and no more failed actions after every commit?

Aprillion:
nope, that didn't help, looks like the `if_merged` part is not syntax, just name of the job and correct syntax is `if: github.event.pull_request.merged == true` 🤦‍♂️

PR #134, opened 2023-01-22
gen encodings from coda questions

ccstan99:
Sorry, I'm not sure what's the best way to add my changes to yours without making a big mess, so I'm parking it here. LMK if I should merged onto `future` branch instead.

- Updated GH actions `gen-encodings-deploy.yml` to pass CODA_TOKEN
- In `questions-encodings.mjs`, use coda to pull UI ID and questions to generate encodings. The JSON file only has the ~216 canonically answered questions, to keep the browser file smaller. I'll encode all ~406 canonical questions including those unanswered in a separate location later. By default, the Coda API only downloads 200 rows, so I set the `limit` to 1000 just to make sure I grab everything.
- In `tfWorker.js`, return the pageid in addition to the question title. I didn't change the component in `search.tsx`. I'll let you decide whether you want to use the pageid or question title.


ccstan99:
Used syntax of https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-when-a-pull-request-merges as guide to add `if: github.event.pull_request.merged == true` but can't test it until it's merged in main branch. I'll try testing it later in forked repo before we merge.

.github/workflows/deploy.yml
```diff
@@ -9,20 +9,20 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone stampy-ui repository code
+        uses: actions/checkout@v3
+      - name: Deploy to Cloudflare
+        run: |
+          cat wrangler.toml.template \
+            | sed s/{CLOUDFLARE_ACCT_ID}/${{ secrets.CLOUDFLARE_ACCT_ID }}/ \
+            | sed s/{STAMPY_KV_ID}/${{ secrets.STAMPY_KV_ID }}/ \
+            | set s/{CODA_TOKEN}/${{ secrets.CODA_TOKEN }} \
```
Aprillion, line 23: aaah, now I see that I forgot `/` at the end of this line, that's why it was complaining about invalid syntax when this change sneaked into the master branch by accident and executed!

I will fix it in both files after merge to `coda-migration`

Aprillion (approved):
Thanks! Looks good, let me merge it and try it right away on my branch.

Simpler workflow would be if we could merge 1 branch to `future` first and then reviewing another PR later, but when both features build on top of each other and both are work in progress at the same time, PRs into feature branches are useful too IMHO 👍

PR #135, opened 2023-01-25
Add new questions when none match

mruwnik:
* fixes bug with sugestions list not being navigable by keyboard

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -2,7 +2,7 @@ name: Deploy to Cloudflare Worker
 on:
   workflow_dispatch:
   pull_request:
-    branches: 
+    branches:
       - master
     types:
       - closed
@@ -21,6 +21,7 @@ jobs:
             | sed s/{CLOUDFLARE_ACCT_ID}/${{ secrets.CLOUDFLARE_ACCT_ID }}/ \
             | sed s/{STAMPY_KV_ID}/${{ secrets.STAMPY_KV_ID }}/ \
             | sed s/{CODA_TOKEN}/${{ secrets.CODA_TOKEN }}/ \
+            | sed s/{CODA_INCOMING_TOKEN}/${{ secrets.CODA_INCOMING_TOKEN }}/ \
             > wrangler.toml
           npm ci
           npm run deploy
--- a/README.md
+++ b/README.md
@@ -31,6 +31,16 @@ Stampy UI is an interface for [stampy.ai](https://stampy.ai/read/Get_involved), 
   (you need access to the doc of course, which you can request on the Discord in source code)
 - Replace the value for `{CODA_TOKEN}` in `wrangler.toml`
 
+  4.1 Setup write access to the "Incoming questions" table in [Coda](https://coda.io/account)
+
+This step is only needed if you want to add new questions to Coda. Seeing as they will then
+go live, it would probably be better to only set this if you are working on that specific
+component, in order to not mess up the list of incoming questions.
+
+- When logged in to Coda, `Generate API token` in your Account settigns
+- Add restrictions: `Doc or table`, `Read and Write`, for the table with url `https://coda.io/d/_dfau7sl2hmG#_tugrid-S_6SYj6Tjm`
+- Replace the value for `{CODA_INCOMING_TOKEN}` in `wrangler.toml`
+
 5. Create stampy-ui Remix environment
 
 - Change directory to where you downloaded the stampy-ui github repository: `cd stampy-ui`
@@ -53,12 +63,13 @@ $ npm run deploy
 Live demo: https://ui.stampy.ai
 
 ## Add new domain
-* log in to [Cloudflare Dashboard](https://dash.cloudflare.com/) owned by @plexish
-* use `Add a site` button on homepage, choose the Free plan
-   * in the DNS section for this site > `Add record` for 2 new CNAME records:
-   * `CNAME   @   stampy-ui.stampy.workers.dev`
-   * `CNAME   *   stampy-ui.stampy.workers.dev`
-* go to Workers > stampy-ui > [Triggers](https://dash.cloudflare.com/841738ad455064a5846675cc41124c85/workers/services/view/stampy-ui/production/triggers)
-   * in Routes section > `Add route` with the new domain (e.g. `example.com/*` and select the zone to the just-added site)
-* [update your domain registrar](https://developers.cloudflare.com/fundamentals/get-started/setup/add-site/#update-your-registrar) to enable CF
-* check the site after a few minutes (CF claims "up to 24 hours", it's usually faster ... but don't share the link too widely on the first day)
+
+- log in to [Cloudflare Dashboard](https://dash.cloudflare.com/) owned by @plexish
+- use `Add a site` button on homepage, choose the Free plan
+  - in the DNS section for this site > `Add record` for 2 new CNAME records:
+  - `CNAME @ stampy-ui.stampy.workers.dev`
+  - `CNAME * stampy-ui.stampy.workers.dev`
+- go to Workers > stampy-ui > [Triggers](https://dash.cloudflare.com/841738ad455064a5846675cc41124c85/workers/services/view/stampy-ui/production/triggers)
+  - in Routes section > `Add route` with the new domain (e.g. `example.com/*` and select the zone to the just-added site)
+- [update your domain registrar](https://developers.cloudflare.com/fundamentals/get-started/setup/add-site/#update-your-registrar) to enable CF
+- check the site after a few minutes (CF claims "up to 24 hours", it's usually faster ... but don't share the link too widely on the first day)
--- a/app/components/search.tsx
+++ b/app/components/search.tsx
@@ -1,6 +1,7 @@
 import {useState, useEffect, useRef, MouseEvent, useMemo, MutableRefObject} from 'react'
 import debounce from 'lodash/debounce'
 import {Question} from '~/routes/questions/$question'
+import {AddQuestion} from '~/routes/questions/add'
 import {MagnifyingGlass} from '~/components/icons-generated'
 import AutoHeight from 'react-auto-height'
 
@@ -85,8 +86,24 @@ export default function Search({
   const results = tfFinishedLoadingRef.current ? searchResults : baselineSearchResults
   const model = tfFinishedLoadingRef.current ? 'tensorflow' : 'plaintext'
 
+  const hideSearchResults = () => setShowResults(false);
+  const handleBlur = (e) => {
+    // If the focus changes from something in the search widget to something outside
+    // of it, then hide the results. If it's just jumping around the results, then keep
+    // them shown.
+    const focusedOnResult = e.relatedTarget?.classList.contains('result-item');
+    setShowResults(focusedOnResult);
+  }
+  const onQuestionAdded = (title) => {
+      alert('Your question "' + title + '" was added. It might take a while for it to be shown here, as it has yet to be answered.');
+      hideSearchResults();
+  }
+
   return (
-    <div>
+    <div
+      onFocus={() => setShowResults(true)}
+      onBlur={handleBlur}
+      >
       <label className="searchbar">
         <input
           type="search"
@@ -95,8 +112,6 @@ export default function Search({
           autoComplete="off"
           onChange={(e) => handleChange(e.currentTarget.value)}
           onKeyDown={(e) => e.key === 'Enter' && searchFn(e.currentTarget.value)}
-          onFocus={() => setShowResults(true)}
-          onBlur={() => setShowResults(false)} // TODO: figure out accessibility - do not blur on keyboard navigation into the result list
         />
         <MagnifyingGlass />
       </label>
@@ -115,22 +130,17 @@ export default function Search({
                     title,
                     score,
                     model,
-                    onSelect,
+                    onSelect: (e) => hideSearchResults() || onSelect(e),
                     isAlreadyOpen: openQuestionTitles.includes(title),
                   }}
                 />
               ))}
           </div>
-          <a
-            href={`https://stampy.ai/wiki/Special:FormStart?form=Q&page_name=${searchInputRef.current}`}
-            target="_blank"
-            rel="noreferrer"
-            className="result-item none-of-the-above"
-            title="Request a new question"
-            onMouseDown={(e) => e.preventDefault()} // prevent onBlur handler before click on the link happens
-          >
-            ＋ None of these: Request an answer to my exact question above
-          </a>
+          <AddQuestion
+              title={searchInputRef.current}
+              relatedQuestions={results.map(({title}) => title)}
+              onQuestionAdded={onQuestionAdded}
+          />
         </div>
       </AutoHeight>
     </div>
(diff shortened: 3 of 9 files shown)


README.md
```diff
@@ -31,6 +31,16 @@ Stampy UI is an interface for [stampy.ai](https://stampy.ai/read/Get_involved),
...
 - Replace the value for `{CODA_TOKEN}` in `wrangler.toml`
 
+4.1 Setup write access to the "Incoming questions" table in [Coda](https://coda.io/account)
+
+This step is only needed if you want to add new questions to Coda. Seeing as they will then
+go live, it would probably be better to only set this if you are working on that specific
+component, in order to not mess up the list of incoming questions.
+
+- When logged in to Coda, `Generate API token` in your Account settigns
+- Add restrictions: `Doc or table`, `Read and Write`, for the table with url `https://coda.io/d/_dfau7sl2hmG#_tugrid-S_6SYj6Tjm`
```
mruwnik, line 41: this really only needs write access. 
I did separate tokens, so people without write access to coda can also develop stuff. Also because I didn't see a way to grant read access to everything, but write only to this one table

app/components/search.tsx
```diff
@@ -85,8 +86,26 @@ export default function Search({
...
+    // React will recreate the whole search widget if `showResults` changes.
+    // `true -> true` also counts as a change, so make sure that it only
+    // gets mutated when it really should be opened/closed
+    const focusedOnResult = e.relatedTarget?.classList.contains('result-item');
+    if (!focusedOnResult) {
+      hideSearchResults();
+    }
+  }
+  const onQuestionAdded = (title) => {
+      alert('Your question "' + title + '" was added. It might take a while for it to be shown here, as it has yet to be answered.');
```
mruwnik, line 100: a better text here would be appreciated.
I did it so this function gets passed into the AddQuestion widget because it needs to hide the list of suggested answers after the new one gets added. I wanted to have a single `hideSearchResults()` function to keep things tidy, which would require passing it in anyway, or having some other mechanism in a global function. This seems like the least ugly.

app/components/search.tsx
```diff
@@ -85,8 +86,26 @@ export default function Search({
...
+      hideSearchResults();
+  }
+
   return (
-    <div>
+    <div
+      onFocus={() => setShowResults(true)}
+      onBlur={handleBlur}
```
mruwnik, line 107: this has to be here for focus changes in the suggested answers to be handled properly - otherwise it would never close once the input loses focus

app/root.css
```diff
@@ -488,6 +488,9 @@ footer > *:not(:last-child) {
   color: inherit;
   font-style: italic;
 }
+.result-item.none-of-the-above button {
+  all: unset; /* Make the button look like a span */
+}
```
mruwnik, line 493: for lack of a better solution and because my css skills are rudimentary at best

app/routes/questions/add.tsx
```diff
@@ -0,0 +1,74 @@
...
+    // Make sure that the question was provided
+    if (!title) return redirect(redirectTo);
+
+    // Check whether the question is simply a prefix of an existant question,
+    // and if so ignore it
+    const allQuestions = await loadAllQuestions(request);
+    const isPrefix = allQuestions.data.some(
+        question => question.toLowerCase().startsWith(title.toLowerCase())
+    );
+    if (isPrefix) return redirect(redirectTo);
```
mruwnik, line 24: I'm not sure about this. It will handle things like "Why might a superintelligent AI be da", but e.g. also remove questions that are valid subsets of existing ones, like "What is AI" vs "What is AI safety"

app/routes/questions/add.tsx
```diff
@@ -0,0 +1,74 @@
...
+export const AddQuestion = ({ title, relatedQuestions, onQuestionAdded }) => {
+    const actionData = useActionData<typeof action>();
+    const [remixSearchParams] = useSearchParams();
+    const [stateString] = useState(() => remixSearchParams.get('state'));
+
+    const transition = useTransition();
+    const submit = useSubmit();
+
+    const handleSubmit = (e) => {
+        e.preventDefault();
```
mruwnik, line 45: this is needed to avoid race conditions between the blur event destroying this widget and the form actually getting submitted - if this function finishes too fast, the form doesn't always get sent

app/routes/questions/add.tsx
```diff
@@ -0,0 +1,74 @@
...
+            <input type="hidden" name="stateString" value={stateString}/>
+            {relatedQuestions.map(title => (
+                <input type="hidden"
+                       name="relatedQuestion"
+                       key={title}
+                       value={title}
+                />
+            ))}
+           {transition.state === "submitting"
+               ? "Adding question..."
```
mruwnik, line 67: this isn't really needed, as submitting this form also results in its destruction

app/server-utils/stampy.ts
```diff
@@ -52,16 +58,17 @@ const CODA_DOC_ID = 'fau7sl2hmG'
...
-  const url = `https://coda.io/apis/v1/docs/${CODA_DOC_ID}/tables/${enc(table)}/rows?${params}`
 
+
+/**
+   Coda has limits on how many rows can be returned at once (default is 200).
+   This function will keep on fetching pages until it has downloaded all items
+   that match the given url.
+   Any errors will cause the whole chain to abort.
+ **/
+const paginateCoda = async (
```
mruwnik, line 69: this is just the body of the `getCodaRows` extracted + a bit of recursion

app/server-utils/stampy.ts
```diff
@@ -115,3 +137,55 @@ export const loadAllCanonicallyAnsweredQuestions = withCache(
...
+export const insertRows = async (
+  table: string,
+  rows: NewQuestion[],
+) => {
+  const url = `https://coda.io/apis/v1/docs/${CODA_DOC_ID}/tables/${enc(table)}/rows`
+  const payload = {
+    'rows': rows.map(row => (
+        {
+          'cells': [
+            { 'column': 'c-Z1ouuaH5zW', 'value': row.title },
```
mruwnik, line 159: would these be better as constants? Or a helper function?

.github/workflows/deploy.yml
```diff
@@ -21,6 +21,7 @@ jobs:
             | sed s/{CLOUDFLARE_ACCT_ID}/${{ secrets.CLOUDFLARE_ACCT_ID }}/ \
             | sed s/{STAMPY_KV_ID}/${{ secrets.STAMPY_KV_ID }}/ \
             | sed s/{CODA_TOKEN}/${{ secrets.CODA_TOKEN }}/ \
+            | sed s/{CODA_INCOMING_TOKEN}/${{ secrets.CODA_INCOMING_TOKEN }}/ \
```
mruwnik, line 24: this will, of course, still need to be created

app/root.css
Aprillion: can you reuse existing class for `.transparent-button`, please?

app/routes/questions/add.tsx
Aprillion: whatever you decide is the desired behaviour in discussion with @plexish about all the branches of this code...

I don't understand all these redirects and haven't seen written specification for this feature, but I don't remember anything about redirects myself 🤔

app/routes/questions/add.tsx
Aprillion: we shouldn't destroy this while loading, can you please keep it visible (don't blur / change detection of blur to keep the search form visible while interacting with the links, please)

app/server-utils/stampy.ts
```diff
@@ -1,15 +1,21 @@
-import {withCache} from '~/server-utils/kv-cache'
-import {Converter} from 'showdown'
+import { withCache } from '~/server-utils/kv-cache'
+import { Converter } from 'showdown'
```
Aprillion, line 2: this code style doesn't look like what we set up in `prettier` - can you enable format on save in your editor and/or  `npm run lint` from command line, please?

app/server-utils/stampy.ts
Aprillion: is it possible to insert rows via API by column name instead of column id?

if not, we should probably have the ids in constants as you suggest,, but not just for writing, also for the GET request (to use the same mechanism and not 2 unsynchronized ways to map columns)

wrangler.toml.template
```diff
@@ -17,3 +17,4 @@ command = ""
 
 [vars]
 CODA_TOKEN = "{CODA_TOKEN}"
+CODA_INCOMING_TOKEN = "{CODA_INCOMING_TOKEN}"
```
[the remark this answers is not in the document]
mruwnik: while generating an API token I can only provide 1 doc or table. A R+W for the whole document would work, but would be less secure than R for the doc and W for this specific table. Dunno how much of an issue that would be

app/routes/questions/add.tsx
mruwnik: Won't that mean that after clicking on this, the user won't see any action until the question gets added to Coda? Or do you mean that adding a question shouldn't close the suggested answers box?

app/routes/questions/add.tsx
mruwnik: Remix tries to gracefully fallback when there is no JS. Which means that action function have to return valid HTTP. This endpoint is `questions/add`, so it needs to make sure that the page stays at `http://127.0.0.1:8787/` after the question is added, but a form submit pretty much is a page reload. So this function needs to both redirect to the root site, but also to make sure that any selected questions (i.e. the state query param) stay selected.
That's a long winded way of saying that you should just think of `redirect(redirectTo)` as a fancy `return`

app/components/search.tsx
```diff
@@ -85,8 +86,26 @@ export default function Search({
   const results = tfFinishedLoadingRef.current ? searchResults : baselineSearchResults
   const model = tfFinishedLoadingRef.current ? 'tensorflow' : 'plaintext'
 
+  const hideSearchResults = () => setShowResults(false);
+  const handleBlur = (e) => {
+    // React will recreate the whole search widget if `showResults` changes.
+    // `true -> true` also counts as a change, so make sure that it only
+    // gets mutated when it really should be opened/closed
```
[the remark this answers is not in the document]
mruwnik: it shouldn't, but when experimenting it did :/ I'll dig in a bit deeper to be sure.
It handles opening in the sense that it sets `setShowResults(false)` (which closes the search widget). So it doesn't really :D The previous blurring problems were caused by the search input losing focus when a suggested question gained focus, so it was setting `setShowResults(false)` and then closing the search widget. This handles the blur by only closing the search widget when focus changes to an element that is outside of the wrapping div, which allows focus to jump around the suggested questions

app/root.css
mruwnik: perfect! Just what I was looking for - added

app/routes/questions/add.tsx
mruwnik: I removed the transition stuff

app/server-utils/stampy.ts
mruwnik: turns out that you can use column names. As long as no one changes them. Should be fine

app/components/search.tsx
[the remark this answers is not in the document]
mruwnik: turns out you're totally correct. Changed this around

README.md
```diff
@@ -53,12 +63,13 @@ $ npm run deploy
...
-* use `Add a site` button on homepage, choose the Free plan
-   * in the DNS section for this site > `Add record` for 2 new CNAME records:
-   * `CNAME   @   stampy-ui.stampy.workers.dev`
-   * `CNAME   *   stampy-ui.stampy.workers.dev`
-* go to Workers > stampy-ui > [Triggers](https://dash.cloudflare.com/841738ad455064a5846675cc41124c85/workers/services/view/stampy-ui/production/triggers)
-   * in Routes section > `Add route` with the new domain (e.g. `example.com/*` and select the zone to the just-added site)
-* [update your domain registrar](https://developers.cloudflare.com/fundamentals/get-started/setup/add-site/#update-your-registrar) to enable CF
-* check the site after a few minutes (CF claims "up to 24 hours", it's usually faster ... but don't share the link too widely on the first day)
+
+- log in to [Cloudflare Dashboard](https://dash.cloudflare.com/) owned by @plexish
```
mruwnik, line 67: `npm run lint` did this

app/routes/questions/add.tsx
Aprillion: ah, ok then 👍

.github/workflows/deploy.yml
Aprillion: - [ ] TODO: create GH secret with this key

README.md
Aprillion: hm, I guess I forgot to configure my editor on the personal computer to run prettier also on markdown files..

app/routes/questions/add.tsx
```diff
@@ -0,0 +1,70 @@
+import {useState} from 'react'
+import type { ActionArgs } from '@remix-run/cloudflare'
```
Aprillion, line 2: hm, format of these lines is still inconsistent,, maybe `lint` script in our package.json is not set up correctly 🤔

I can investigate after this PR

Aprillion (approved):

app/components/search.tsx
plexish: 'Thanks for asking a new question! "' + title + '" was added to our [suggestion box](https://coda.io/@alignmentdev/ai-safety-info/suggested-questions-66). It might take a while for it to be answered by our writers, but check back in a few months.'

app/routes/questions/add.tsx
plexish: > "What is AI" vs "What is AI safety"
hm, this seems like a small enough set of edge cases that's it's not too bad, but also adding substrings is not a big problem. Not a strong preference here, whichever is easy/already done seems fine.

app/components/search.tsx
```diff
@@ -85,8 +86,28 @@ export default function Search({
...
+    // them shown.
+    const focusedOnResult = e.relatedTarget?.classList.contains('result-item');
+    setShowResults(focusedOnResult);
+  }
+  const onQuestionAdded = (title: string) => {
+      hideSearchResults();
+      alert(
+          'Thanks for asking a new question! "' + title + '" was added to our suggestion box ' +
+          'It might take a while for it to be answered by our writers, but check back in a few months.\n\n' +
+          'The list of current suggestions can be found at https://coda.io/@alignmentdev/ai-safety-info/suggested-questions-66'
```
mruwnik, line 102: alerts don't allow html, so I just did it manually here

app/routes/questions/add.tsx
```diff
@@ -0,0 +1,86 @@
...
+    // Check whether the question is simply a prefix of an existant question,
+    // and if so ignore it
+    const allQuestions = await loadAllQuestions(request);
+    const isPrefix = allQuestions.data.some(
+        question => question.toLowerCase().startsWith(title.toLowerCase())
+    );
+    if (isPrefix) return redirect(redirectTo);
+
+    // Try to get the related questions from semantic search, and if that doesn't work,
+    // fallback to sending along whatever was displayed by the search box
```
mruwnik, line 34: get related questions from semantic search

wrangler.toml.template
```diff
@@ -17,3 +17,5 @@ command = ""
...
+NLP_SEARCH_ENDPOINT = "https://stampy-nlp.uw.r.appspot.com"
```
mruwnik, line 21: I thought it better to configure urls etc. that can change via wrangler, rather than having them directly in the code - should I change it back?

app/server-utils/stampy.ts
```diff
@@ -50,32 +56,70 @@ type CodaRow = {
 
 const CODA_DOC_ID = 'fau7sl2hmG'
 
+// Use table ide, rather than names, in case of renames
```
mruwnik, line 59: tables were getting renamed, so I moved to ids

Aprillion (approved):
cool, feel free to merge if it works, I will check the prettier settings during the weekend

PR #136, opened 2023-01-28
Fix typescript, eslint, prettier and change UI ID field type

Aprillion:


app/routes/questions/add.tsx
```diff
@@ -1,86 +1,112 @@
...
-                ＋ None of these: Request an answer to my exact question above
-            </button>
-        </Form>
+      <p className="result-item none-of-the-above">
+        Thanks for asking a new question! {title} was added to our suggestion box. It might take a
+        while for it to be answered by our writers, but check back in a few months.
+        <br />
+        <br />
+        The list of current suggestions can be found at{' '}
+        <a href="https://coda.io/@alignmentdev/ai-safety-info/suggested-questions-66">
```
mruwnik, line 87: this link could be moved into the previous text, e.g. "added to our <a>suggestion box</a>".
I only did it this way because alerts don't allow html

mruwnik (approved):

app/routes/questions/add.tsx
Aprillion: can be changed on hackathon 👍

PR #138, opened 2023-01-30
Hackathon

Aprillion:
Changes from online Discord collab.


app/hooks/useQuestionStateInUrl.ts
```diff
@@ -114,6 +124,7 @@ export default function useQuestionStateInUrl(minLogo: boolean, initialQuestions
           return `${k}${v}`
         })
         .join('')
+      console.log(pageid, currentState, newState)
```
mruwnik, line 127: ?

app/hooks/useQuestionStateInUrl.ts
```diff
@@ -82,29 +90,31 @@ export default function useQuestionStateInUrl(minLogo: boolean, initialQuestions
...
-        const isAlreadyDisplayed = questions.some(({title}) => title === q.title)
-        return hasCanonicalAnswer && !isAlreadyDisplayed
+        const isAlreadyDisplayed = questions.some(({pageid}) => pageid === question.pageid)
+        return isOnSite && !isAlreadyDisplayed
       })
 
       const newState = getStateEntries(currentState)
         .map(([k, v]) => {
-          if (k === pageid) {
+          if (k === pageid.toString()) {
```
mruwnik, line 117: isn't pageid always a string?

app/routes/index.tsx
```diff
@@ -53,6 +54,18 @@ export default function App() {
     .filter(({questionState}) => questionState === '_')
     .map(({title}) => title)
 
+  const handleSpecialLinks = (e: MouseEvent) => {
+    const el = e.target as HTMLAnchorElement
+    if (el.tagName !== 'A') return
```
mruwnik, line 59: is this always uppercase?

mruwnik (approved):

app/hooks/useQuestionStateInUrl.ts
Aprillion: aaah, forgotten debug, thanks 😅

app/hooks/useQuestionStateInUrl.ts
Aprillion: should be after we re-generate tensorflow embeddings, but I figured if Coda doesn't guarantee the type of every value in a column, might be safer to keep this just in case future embeddigns will end up with accidental numbers too

app/routes/index.tsx
Aprillion: yes, https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName

mruwnik (approved):

PR #143, opened 2023-02-04
Future

Aprillion:
@plexish gave a green light for switching to Coda, which should deploy to `aisafety.info` by GH action automatically by merging to master 🤞 (I created all GH secrets already when updating the actions files, but it might be wise to keep an eye on everything after merge).

Current version manually deployed to https://stampy-ui.aprillion.workers.dev for final testing (including direct link to `Dashboard` in the footer from latest discussion on Discord).


app/root.css
```diff
@@ -345,11 +337,13 @@ tr:nth-child(even) {
...
-.answer.short .card-longdesc,
-.answer.short .card-hide-longdesc {
-  display: none;
+.answer img {
+  max-width: 100%;
+}
+
+.answer iframe {
+  width: 100%;
+  aspect-ratio: 16/9;
```
ChrisRimmer, line 346: :exploding_head:

questions-encodings.mjs
```diff
@@ -33,14 +18,27 @@ if (fs.existsSync(filename)) {
...
 
-const questions = response.map(({fulltext}) => fulltext)
+const CODA_TOKEN = process.argv[2]
+const CODA_DOC_ID = 'fau7sl2hmG'
+const table = 'Answers' 
+const params = 'useColumnNames=true&query=Status:%22Live%20on%20site%22&limit=1000' 
+const url =
+  `https://coda.io/apis/v1/docs/${CODA_DOC_ID}/tables/${table}/rows?${params}`
+
+const r = await (await fetch(url, {headers: {Authorization: `Bearer ${CODA_TOKEN}`}})).json()
```
ChrisRimmer, line 34: Very relieved I'm not the only person who does this `await (await fetch()).json()` :D

ChrisRimmer (dismissed):
Would be nice to revisit some of the machinery in here and be a bit more consistent with usage of async/await vs `fetch().then()`, and I think it would be nice to mix in a few more structured programming principles (preferring not to use early returns, preferring `if () {return foo} else {return bar}` over `if () {return foo} return bar` etc) but really I don't see anything here that I want to tap the brakes on. lgtm :shipit:

app/routes/index.tsx
```diff
@@ -112,14 +123,17 @@ export default function App() {
...
+        <a href="https://get_involved.aisafety.info">About / Get Involved</a>
+        <a href="https://dashboard.aisafety.info">Dashboard</a>
         <a href="https://github.com/StampyAI/stampy-ui">Help Code</a>
-        <a href="https://stampy.ai/wiki/Discord_invite">Join Discord</a>
         <a href="https://docs.google.com/forms/d/e/1FAIpQLSdT--8lx5F2pAZoRPPkDusA7vUTvKTVnNiAb9U5cqnohDhzHA/viewform">
           Feedback
         </a>
-        <a href="https://stampy.ai/wiki/Meta:Copyrights">@ 2022 stampy.ai</a>
+        <a href="https://coda.io/d/AI-Safety-Info-Dashboard_dfau7sl2hmG/Copyright_su79L#_luPMa">
+          @ 2022 stampy.ai
```
mruwnik, line 135: this should probably be changed to some kind of auto incrementor or something :D

mruwnik (dismissed):

app/routes/index.tsx
Aprillion: hmm, and we should also use copyright symbol © instead of @

mruwnik (approved):

PR #145, opened 2023-02-07
Display tags for questions

mruwnik:
image

image

### DIFF
--- a/app/components/dialog.tsx
+++ b/app/components/dialog.tsx
@@ -1,0 +1,41 @@
+import {useCallback, useRef} from 'react'
+
+type Props = {
+  children: any
+  onClose: (e: any) => void
+  [k: string]: any
+}
+
+export default function Dialog({children, onClose}: Props) {
+  // React magic to get the actual DOM element once it's been mounted
+  const dialogRef = useRef<HTMLDialogElement>(null)
+  const dialogSetter = useCallback((node: HTMLDialogElement) => {
+    if (node !== null) {
+      // Show the dialog this way, in order for it to automatically handle escapes
+      node.showModal()
+      dialogRef.current = node
+    }
+  }, [])
+
+  // Calculate whether the given click event was outside of the dialog, and if so close it
+  const closeIfOutside = (e: React.MouseEvent<HTMLElement>) => {
+    // If there isn't any dialog displayed, then the click obviously wasn't in it
+    if (dialogRef?.current == null) return
+
+    const rect = dialogRef.current.getBoundingClientRect()
+    const isInDialog =
+      rect.top <= e.clientY &&
+      e.clientY <= rect.top + rect.height &&
+      rect.left <= e.clientX &&
+      e.clientX <= rect.left + rect.width
+    if (!isInDialog) {
+      dialogRef.current.close()
+    }
+  }
+
+  return (
+    <dialog ref={dialogSetter} className="dialog" onClose={onClose} onClick={closeIfOutside}>
+      <div className="dialog-contents">{children}</div>
+    </dialog>
+  )
+}
--- a/app/components/tags.tsx
+++ b/app/components/tags.tsx
@@ -1,0 +1,21 @@
+import type {Tag as TagType} from '~/server-utils/stampy'
+
+type Props = {
+  tags: TagType[]
+  [k: string]: any
+}
+
+export default function Tags({tags}: Props) {
+  return (
+    <div className="tags-container">
+      <div>
+        {tags &&
+          tags.filter(Boolean).map((t) => (
+            <span className="question-tag" key="{t.tagId}">
+              {t.name}
+            </span>
+          ))}
+      </div>
+    </div>
+  )
+}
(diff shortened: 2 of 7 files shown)


app/components/tags.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+type Props = {
+  tags: TagType[]
+  [k: string]: any
+}
+
+export default function Tags({tags}: Props) {
+  return (
+    <div className="tags-container">
+      <div>
```
mruwnik, line 11: this extra div is because I don't do CSS - my solution to having the tags and actions on one line is:
untitled


Please tell me there's a better way...

app/root.css
```diff
@@ -346,6 +346,24 @@ tr:nth-child(even) {
...
+  display: grid;
+}
+.question-tag {
+  outline: 1px solid var(--borderColor);
+  border-radius: 4px;
+  box-shadow: 0 0 8px rgba(0, 0, 0, 0.1);
+  margin-right: 10px;
+  padding: 5px;
+  line-height: 2.3;
+}
```
mruwnik, line 365: adding more CSS, rather than proper classes - Hooray!

app/server-utils/stampy.ts
```diff
@@ -156,6 +174,31 @@ export const loadAllQuestions = withCache('allQuestions', async () => {
...
+      return question
+  }
+}
+const toTag = (r: CodaRow): Tag => ({
+  rowId: r.id,
+  tagId: r.values['Tag ID'],
+  name: r.name,
+  url: r.browserLink,
+  internal: r.values['Internal?'],
+  questions: r.values['Questions tagged with this'].map(({name}) => name),
```
mruwnik, line 193: this is just a list of questions titles, as the query doesn't return the UI IDs - one of the following will need to be done at some point:
* expand this query to also get the UI IDs here
* mix this with `loadAllQuestions()` to make sure the ids are known
* update the tags with the relative ids when an actual question is viewed
* worry about it when someone clicks on a tag

The first 2 solutions will save overall queries/latency at the cost of more stuff needed to be stored on the server. Probably won't be a problem until there are a few million questions... They also make the first loading of the site a bit longer (subsequent ones are fine, as the results are cached)
The other options are nicer in that they're explicit with how they utilise magic, which is always a bonus

app/components/dialog.tsx
```diff
@@ -0,0 +1,41 @@
...
+      e.clientY <= rect.top + rect.height &&
+      rect.left <= e.clientX &&
+      e.clientX <= rect.left + rect.width
+    if (!isInDialog) {
+      dialogRef.current.close()
+    }
+  }
+
+  return (
+    <dialog ref={dialogSetter} className="dialog" onClose={onClose} onClick={closeIfOutside}>
```
mruwnik, line 37: I wanted to use the HTML5 dialog element. It can simply have an 'open' attribute to avoid all this ref magic, but the spec recommends calling `showModal()`, and it also handles escapes etc nicely if done this way

app/server-utils/stampy.ts
```diff
@@ -114,20 +143,27 @@ const getCodaRows = async (
 
 const md = new MarkdownIt({html: true}).use(MarkdownItFootnote)
 
-const extractText = (markdown: string) => markdown?.replace(/^```|```$/g, '')
+// Sometimes string fields are returned as lists. This can happen when there are duplicate entries in Coda
```
mruwnik, line 146: this was fun. The offending coda row is now fixed, but I left this here just in case - otherwise it's enough for someone to create 2 gdocs with the same name to blow up everything

app/server-utils/stampy.ts
```diff
@@ -153,7 +189,38 @@ export const loadOnSiteAnswers = withCache('onSiteAnswers', async () => {
...
+  questions: r.values['Questions tagged with this']
+    .map(({name}) => ({title: name, pageid: nameToId[name]}))
+    .filter((q) => q.pageid),
+  mainQuestion: extractMainQuestion(r.values['Main question']),
+})
+export const loadAllTags = withCache('allTags', async () => {
+  const rows = await getCodaRows(TAGS_TABLE)
+  const questions = await loadAllQuestions({url: ''})
+  const nameToId = Object.fromEntries(
+    questions.data.filter((q) => q.status == QuestionStatus.LIVE).map((q) => [q.title, q.pageid])
```
mruwnik, line 220: I assume that only live questions should be shown in the popup?

app/server-utils/stampy.ts
```diff
@@ -153,7 +189,38 @@ export const loadOnSiteAnswers = withCache('onSiteAnswers', async () => {
...
+  url: r.browserLink,
+  internal: r.values['Internal?'],
+  questions: r.values['Questions tagged with this']
+    .map(({name}) => ({title: name, pageid: nameToId[name]}))
+    .filter((q) => q.pageid),
+  mainQuestion: extractMainQuestion(r.values['Main question']),
+})
+export const loadAllTags = withCache('allTags', async () => {
+  const rows = await getCodaRows(TAGS_TABLE)
+  const questions = await loadAllQuestions({url: ''})
```
mruwnik, line 218: this `{url: ''}` is to use the cacheing mechanism for all the questions. I don't like the added complexity, but it makes things a bit easier

Aprillion (approved):

app/server-utils/stampy.ts
plexish: Yes.

plexish:
Nice work!

PR #146, opened 2023-02-09
Add infini-scroll

mruwnik:
image
I added a loader type thingy, but left the footer for now

### DIFF
--- a/app/components/infiniteScroll.tsx
+++ b/app/components/infiniteScroll.tsx
@@ -5,7 +5,7 @@ type Props = {
   perPage: number
   fetchMore: (page: number, perPage: number) => any[]
   loadingRenderer: (loading: boolean) => ReactNode
-  [k: string]: any
+  [k: string]: JSX.IntrinsicElements['div']
 }
 
 export default function InfiniteScroll({
--- a/app/components/infiniteScroll.tsx
+++ b/app/components/infiniteScroll.tsx
@@ -1,0 +1,60 @@
+import {ReactNode, useState, useRef, useCallback} from 'react'
+
+type Props = {
+  children: ReactNode[]
+  perPage: number
+  fetchMore: (page: number, perPage: number) => any[]
+  loadingRenderer: (loading: boolean) => ReactNode
+  [k: string]: any
+}
+
+export default function InfiniteScroll({
+  children,
+  fetchMore,
+  perPage = 10,
+  loadingRenderer = (loading: boolean) => loading && <div className="loader"></div>,
+  ...props
+}: Props) {
+  const [pageNum, setPageNum] = useState(0)
+  const [hasMore, setHasMore] = useState(true)
+  const [loading, setLoading] = useState(false)
+
+  const getMore = useCallback(async () => {
+    if (loading || !hasMore) return
+
+    setLoading(true)
+
+    const nextPage = await fetchMore(pageNum, perPage)
+    setPageNum(pageNum + 1)
+
+    if (nextPage?.length === 0) {
+      setHasMore(false)
+    }
+
+    setLoading(false)
+  }, [loading, hasMore, pageNum, setLoading, setPageNum, setHasMore, fetchMore, perPage])
+
+  const observer = useRef<IntersectionObserver>(null)
+  const lastElementRef = useCallback(
+    (node) => {
+      if (observer.current) observer.current.disconnect()
+
+      if (node) {
+        observer.current = new IntersectionObserver((entries) => {
+          if (entries[0].isIntersecting) {
+            getMore()
+          }
+        })
+        observer.current.observe(node)
+      }
+    },
+    [getMore]
+  )
+  return (
+    <div {...props}>
+      {children}
+      {loadingRenderer(loading)}
+      <div ref={lastElementRef}></div>
+    </div>
+  )
+}
--- a/app/hooks/useQuestionStateInUrl.ts
+++ b/app/hooks/useQuestionStateInUrl.ts
@@ -1,22 +1,25 @@
 import {useState, useRef, useEffect, useMemo, useCallback} from 'react'
 import type {MouseEvent} from 'react'
 import {useSearchParams, useTransition} from '@remix-run/react'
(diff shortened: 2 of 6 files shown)


app/hooks/useQuestionStateInUrl.ts
```diff
@@ -1,22 +1,25 @@
...
 
-const getStateEntries = (state: string): [string, QuestionState][] =>
+type PageId = Question['pageid']
+
+const getStateEntries = (state: string): [PageId, QuestionState][] =>
   Array.from(state.matchAll(/([^-_r]+)([-_r]*)/g) ?? []).map((groups) => [
-    groups[1],
-    (groups[2] || '_') as QuestionState,
+    groups[1], // question id
+    (groups[2] as QuestionState) || QuestionState.OPEN,
```
mruwnik, line 12: I was finding it hard to remember what char meant what, so I went and made an enum for it

app/hooks/useQuestionStateInUrl.ts
```diff
@@ -83,17 +86,111 @@ export default function useQuestionStateInUrl(minLogo: boolean, initialQuestions
...
+  )
+
+  /*
+   * Add the given `questions` to the global questions map. This will not update the URL
+   */
+  const mergeNewQuestions = useCallback((questions: Question[]) => {
+    setQuestionMap((currentMap) =>
+      questions.reduce((map, question) => updateQuestionMap(question, map), new Map(currentMap))
+    )
+  }, [])
```
mruwnik, line 162: the previous functions were simply extracted from `toggleQuestion` so they could be used elsewhere

app/hooks/useQuestionStateInUrl.ts
```diff
@@ -83,17 +86,111 @@ export default function useQuestionStateInUrl(minLogo: boolean, initialQuestions
...
+
+  /*
+   * Add the given `newQuestions` to the collection of questions. This will also update the URL
+   */
+  const addQuestions = useCallback(
```
mruwnik, line 167: this function is new and is the main real change in this file

Aprillion:
> but left the footer for now

as I wrote in #77 I don't agree with keeping both the infinite scrolling and the footer.

app/server-utils/stampy.ts
```diff
@@ -2,7 +2,12 @@ import {withCache} from '~/server-utils/kv-cache'
 import MarkdownIt from 'markdown-it'
 import MarkdownItFootnote from 'markdown-it-footnote'
 
-export type QuestionState = '_' | '-' | 'r'
+// export type QuestionState = '_' | '-' | 'r'
```
Aprillion, line 5: if we don't need to keep commented-out code for some special reason, please delete.

app/server-utils/stampy.ts
```diff
@@ -2,7 +2,12 @@ import {withCache} from '~/server-utils/kv-cache'
...
+export enum QuestionState {
+  OPEN = '_',
+  COLLAPSED = '-',
+  SUBQUESTION = 'r',
```
Aprillion, line 9: `r` was a short for RELATED question

app/root.css
```diff
@@ -26,6 +26,25 @@
   /* dark theme at the end of file to overwrite all previous rules */
 }
 
+.loader {
+  border: 4px solid #f3f3f3; /* Light grey */
+  border-top: 4px solid #3498db; /* Blue */
```
Aprillion, line 31: do both colors work well in both light and dark theme?

app/components/infiniteScroll.tsx
```diff
@@ -0,0 +1,60 @@
+import {ReactNode, useState, useRef, useCallback} from 'react'
+
+type Props = {
+  children: ReactNode[]
+  perPage: number
+  fetchMore: (page: number, perPage: number) => any[]
+  loadingRenderer: (loading: boolean) => ReactNode
+  [k: string]: any
```
Aprillion, line 8: is there some non-any type we could use, please? e.g. `JSX.IntrinsicElements['div']` or `React.HTMLAttributes<HTMLDivElement>`?

app/server-utils/stampy.ts
mruwnik: whoops - good catch

app/components/infiniteScroll.tsx
mruwnik: nice!

app/root.css
mruwnik: image
Depends on what you want :) To be honest, this was pretty much the first css loader that I found, so any corrections/enhancements are welcome

app/server-utils/stampy.ts
mruwnik: updated

app/root.css
Aprillion: yeah, visibe enough for now, can be improved later if someone will find a better color

Aprillion (approved):
Should be good-enough to merge with the floating join discord icon, `Dashboard` and `Feedback` are less critical to figure out .. as long as @plexish or someone is keeping track what needs to move where.

Leaving copyright in footer despite infinite scrolling should be fine, no user reads it and developers can check inside code.

plexish:
Feedback got two responses total despite me promoting it, and Dashboard is.. not essential, it's mostly useful for reviewers who should be able to find their way there anyway as they're meant to be familiar with the system and it's only one extra click from Get Involved. This can go in now that there's a Discord link, thanks for making it!

#77 mentioned using wiki data (sorry if you followed that dated spec!), but as the wiki's going away using coda is the way to go. Using the ordering on Initial Answers (allowing non-Integer values, so it's easier to re-order), and using the variable formula in that to determine how many to display initially.

mruwnik:
Currently it loads all the questions from the `initial questions` page, and if you then scroll down, will fetch another 10 questions from the `On-site API` table, skipping any questions that have already been displayed. This can result in slightly fewer than 10 questions per load, but I'm guessing that's fine :D

plexish:
Okay, that seems reasonable. I can make `On-site API` sorted by something sensible later.

Aprillion (approved):

Issue #140, opened 2023-01-31
Whitelist of plain URLs to iframe


plexish:
Would be super nice to have https://aisafety.world/map iframed on the main page at least, maybe https://coda.io/@alignmentdev/alignmentecosystemdevelopment in another answer, having that list expandable might prove valuable if not too complex.

If a full plain URL from the whitelisted domains is in the markdown, it should be transformed into a large (~screen height, full answer width) iframe of that URL.

- [ ] Airtable
- [ ] Coda
- [ ] https://aisafety.world/

jrhender:
I'm going to take a look at this as a part of the EA London Hackathon 🚀

jrhender:
Here's an example:
<img width="592" alt="image" src="https://user-images.githubusercontent.com/4407464/218254771-09341fa7-ee5b-4ce1-af7d-9e327d2ac567.png">

In the above, https://google.com should render as an iframe of https://google.com

jrhender:
Another reason for having the whitelist (as mentioned in the issue description) is that it might be good to test if sites have set X-Frame-Options such that their site can be loaded in an iframe: https://stackoverflow.com/questions/27358966/how-to-set-x-frame-options-on-iframe

jrhender:
Discussing with @plexish , the "parser" tag (currently applied to this issue) may be inaccurate as it implies that the url -> iframe transformation should be done prior to the data being served to the frontend from coda. This is not ideal because the coda data is also consumed by Discord which cannot render iframes.

PR #148, opened 2023-02-11
#140 / url to iframe

jrhender:
Implementation of URLs in answers rendered in an iframe

For #140 

Todo:
- [x] whitelist of sites. Need to match on domain, not URL as it is the domain that is known, not the URL.
- [x] test each of the whitelisted URLs (find some example questions that have them)
- [x] test that it works for multiple urls in the same text
- [x] fix loading of questions after the question with the iframe -> just needed to rebase
- [x] check security of iframe -> `sandbox` values of iframe should be set per host
- [ ] parse markdown instead of html


app/routes/questions/$question.tsx
```diff
@@ -102,7 +102,7 @@ export function Question({
   } else if (text == null) {
     html = 'Loading...'
   } else {
-    html = text
+    html = text;
```
Aprillion, line 105: don't forget to run prettier with project settings please, not the defaults.

jrhender: Thanks!

jrhender:
Working for question https://aisafety.info/?state=8392_

image

Running locally:

image

jrhender:
Hmm, not quite working still. The loading of questions below the question with the iframe is not working.

plexish:
@jrhender does the iframe have scroll bars?

jrhender:
@plexish I'm able to scroll with my mac trackpad: https://www.loom.com/share/3ff6c6d048a24e66819b5de02d52f4f6
but I'm not sure how it would behave with a mouse. That might need to be tested.

plexish:
Okay, yep, I see a scroll thingy, should work fine :)

app/server-utils/url-to-iframe.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Replaces all of the URLs in text with iframes of the URLs
+ * @param text text in which URLs should be replaced
+ */
+export const urlToIframe = (text: string): string => {
+  const whitelistedHosts: Record<string, HostConfig> = {
+    'aisafety.world': {sandboxValue: 'allow-scripts allow-same-origin'},
+  }
+
+  // Regex is from: https://stackoverflow.com/a/26764609
+  const anchorTagRegex = new RegExp(/<a[\s]+([^>]+)>((?:.(?!<\/a>))*.)<\/a>/gi)
```
mruwnik, line 11: something like `<a rel="alternate" href="https://aisafety.world/" target="_blank">https://aisafety.world/</a>` will break this. A simple fix would be to change the regex to `<a\s+[^>]*?(href="[^>]+")[^>]*?>((?:.(?!<\/a>))*.)<\/a>`. That being said, regexing HTML is generally a bad idea. [This](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454#1732454) is the classic reference.

A more generic option would be to use DOMParser, e.g.:
```
   const parser = new DOMParser();
    const htmlDoc = parser.parseFromString(text, 'text/html');
    for(const a of htmlDoc.getElementsByTagName('a')) {
        const host = new URL(a.href).host
        const hostConfig = whitelistedHosts[host]
        if (hostConfig && a.href.trim() == a.innerHTML.trim()) {
            a.outerHTML = `<iframe src="${a.href}" sandbox="${hostConfig.sandboxValue}"></iframe>`
        }
    }
    return htmlDoc.body.innerHTML
```
The main problem with this, of course, is that it will only work in the browser. So you'd have to move the call from `stampy.ts` to `routes/questions/$question.tsx`. The thing with that is Remix calls routes both in the browser (which is what you want) and on the backend (which will cause an error). There are various ways to handle this, the simplest being just wrapping the parser like this:
```
    let parser
    try {
        parser = new DOMParser();
    } catch (error) {
        return text
    }
```

A nice thing about your solution is that it's cached, so the transformation is called once per question, per API call. Moving this to the frontend would break that, and parsing is generally a lot slower that a basic regex. That being said, unless the questions get too bulky, it shouldn't be a problem?

app/server-utils/stampy.ts
```diff
@@ -1,4 +1,5 @@
 import {withCache} from '~/server-utils/kv-cache'
+import {urlToIframe} from './url-to-iframe'
```
mruwnik, line 2: could you change this to `'~/server-utils/url-to-iframe'`, for consistency?

mruwnik:
Nice result :D I like how the map looks with this

app/server-utils/url-to-iframe.ts
```diff
@@ -0,0 +1,35 @@
...
+  let updatedText = text
+  for (const match of matches) {
+    // Example fulltag: '<a href="https://aisafety.world/">https://aisafety.world/</a>'
+    const fullTag = match[0]
+    // Example hrefUrl: 'https://aisafety.world/'
+    const hrefUrl = match[1].replace('href=', '').replace(/"/g, '')
+    const host = new URL(hrefUrl).host
+    // Example tagContent: 'https://aisafety.world/'
+    const tagContent = match[2]
+    if (Object.keys(whitelistedHosts).includes(host) && hrefUrl === tagContent) {
```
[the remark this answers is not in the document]
jrhender: Great point 👍

app/server-utils/stampy.ts
jrhender: Ah yes, certainly, thanks for pointing that out.

app/server-utils/url-to-iframe.ts
mruwnik: That being said, regex on markdown would be fine. How about moving the regexes earlier, before the contents get transformed from markdown to HTML?

app/server-utils/url-to-iframe.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * Replaces all of the URLs in text with iframes of the URLs
+ * @param text text in which URLs should be replaced
+ */
+export const urlToIframe = (text: string): string => {
+  const whitelistedHosts: Record<string, HostConfig> = {
+    'aisafety.world': {sandboxValue: 'allow-scripts allow-same-origin'},
```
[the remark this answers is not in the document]
jrhender: From my testing:
- `allow-scripts` is needed for the animation when mousing over an element on the map and I think for AJAX to get data
- `allow-same-origin` is needed for the AJAX requests, I believe

Without allowing the above, the map appears but no items are displayed:
image

app/server-utils/url-to-iframe.ts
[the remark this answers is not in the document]
jrhender: Updated in https://github.com/StampyAI/stampy-ui/pull/148/commits/ae0479391c77a06b4ca557d888acc65c265011e7

app/server-utils/stampy.ts
jrhender: Fixed in https://github.com/StampyAI/stampy-ui/pull/148/commits/dd14d29b973ee23221cbe391c9e1c1535d4653c1

app/server-utils/url-to-iframe.ts
jrhender: > Remix calls routes both in the browser (which is what you want) and on the backend (which will cause an error)

Is this referring to serverside rendering from Remix?

app/server-utils/stampy.ts
```diff
@@ -174,7 +175,7 @@ const extractLink = (markdown: string) => markdown?.replace(/^.*\(|\)/g, '')
 const convertToQuestion = (title: string, v: CodaRow['values']): Question => ({
   title,
   pageid: extractText(v['UI ID']),
-  text: v['Rich Text'] ? md.render(extractText(v['Rich Text'])) : null,
+  text: v['Rich Text'] ? urlToIframe(md.render(extractText(v['Rich Text']))) : null,
```
mruwnik, line 178: this doesn't look right - the previous version should be `text: renderText(v['Rich Text']),` - are you sure you're up to date?

app/server-utils/url-to-iframe.ts
mruwnik: yes. If you call this from `stampy.ts::renderText()`, you'll be able to play with markdown, rather than html, which is a lot simpler

app/server-utils/stampy.ts
```diff
@@ -211,21 +211,25 @@ const head = (item: any) => {
...
-      ? v['Related Answers'].map(({name}, i) => ({
-          title: name,
-          pageid: extractText(v['Related IDs'][i]),
-        }))
-      : [],
-  status: v['Status']?.name as QuestionStatus,
-})
+const convertToQuestion = (title: string, v: CodaRow['values']): Question => {
+  const renderedText = renderText(v['Rich Text'])
+  const text = renderedText ? urlToIframe(renderedText) : renderedText
```
mruwnik, line 216: Could you move the `urlToIframe` into `renderText`, please?

app/server-utils/stampy.ts
jrhender: You were right; it wasn't upto date. I had to merge in latest `master` branch

jrhender: Fixed with https://github.com/StampyAI/stampy-ui/pull/148/commits/3af60d3f7db40d6332f1273b9476129ae9c99589

app/server-utils/stampy.ts
jrhender: Yes, great idea, I will do that 💡

mruwnik (approved):

app/server-utils/stampy.ts
jrhender: Implemented in https://github.com/StampyAI/stampy-ui/commit/a679fb17af839571520a946d3486f35aae6d5ca2

mruwnik (approved):

mruwnik:
looks good - merging

app/server-utils/url-to-iframe.ts
jrhender: @mruwnik I've been thinking about this a bit more and I'm not sure parsing the markdown is better than parsing the html. The main reason is that there is no mapping in markdown to an iframe (which is what we need to get to)
In other words:
- in html, the mapping is from `<a>` tag to `<iframe>`
- in markdown, the mapping would be from `[]()` links to ???

Maybe one could
1. find the relevant links in the markdown
2. replace with a special symbol
3. convert this special symbol to `iframe`

Or do you see more clearly how it would work in markdown?

I mean, I see this is merged already, so maybe not a big deal

jrhender:
> something like <a rel="alternate" href="https://aisafety.world/" target="_blank">https://aisafety.world/</a> will break this. A simple fix would be to change the regex to <a\s+[^>]*?(href="[^>]+")[^>]*?>((?:.(?!<\/a>))*.)<\/a>. 

Might still need to handle the above but can do in a separate PR I suppose.

app/server-utils/url-to-iframe.ts
mruwnik: I'm guessing you could do pretty much exactly the same as you do here, but the regex would be simpler. Though I don't know if the HTML generator would correctly handle embedded HTML. Could be checked if you're feeling bored...

Issue #150, opened 2023-02-11
Answers with an alphanumeric rather than numeric ID show permanent loading screen


plexish:
Answers on [this page](https://coda.io/d/AI-Safety-Info_dfau7sl2hmG/Admin-Panel_su93i#Answers_tuile/r452) e.g. [https://aisafety.info/?state=85E0_](https://aisafety.info/?state=85E0_)

PR #151, opened 2023-02-11
fix loading by alphanumeric question id #150

Aprillion:
image

### DIFF
--- a/app/server-utils/stampy.ts
+++ b/app/server-utils/stampy.ts
@@ -108,7 +108,7 @@ export const fetchJson = async (url: string, params?: Record<string, any>) => {
 export const fetchJsonList = async (url: string, params?: Record<string, any>) => {
   const json = await fetchJson(url, params)
   if (!json.items || json.items.length === 0) {
-    throw Error('Empty response')
+    throw Error(`Empty response for ${url}`)
   }
   return json
 }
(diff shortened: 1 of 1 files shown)


ChrisRimmer (approved):
Lgtm ship it

app/server-utils/stampy.ts
```diff
@@ -169,7 +169,9 @@ const convertToQuestion = (title: string, v: CodaRow['values']): Question => ({
 export const loadQuestionDetail = withCache('questionDetail', async (question: string) => {
   const rows = await getCodaRows(
     QUESTION_DETAILS_TABLE,
-    question.match(/^\d+$/) ? 'UI ID' : 'Name',
+    // ids are now alphanumerical, so not possible to detect id by regex match for \d
+    // let's detect ids by length, hopefully no one will make question name so short
+    question.length <= 6 ? 'UI ID' : 'Name',
```
mruwnik, line 174: perhaps `/^[\dA-Z]{3,5}$/` would be safer? Though this should be fine (unless it starts getting weird characters...)

mruwnik (approved):

app/server-utils/stampy.ts
Aprillion: nah, removing the match by Name would be best, but I am not sure if any part of the UI still needs it (I would hope not, but I'm a bit tired right now to check everything - feel free to check and if not needed, make a PR to remove match by `Name` 🤞)

Issue #51, opened 2022-07-04
Tooltips background color


ccstan99:
All tooltips (icons at top, actions on questions, search question scores) should have the same background color #fed.

ChrisRimmer:
Strictly speaking I wonder if we should avoid putting match scores in tooltips - at least if we're using the browser's built-in alt-text. I'm not super duper familiar with screen readers (you'd think after doing this almost a decade I would be, but alas!) but I'm fairly sure the title attributes we're using to generate these tooltips should just contain informative text for screen readers or other accessibility aids.

If we want to surface this info we can (are relevance scores useful? I want to say not very but I could easily be persuaded), but I think it would be most right to develop a little info tooltip component, which we can flag to the browser as explicitly not being an accessibility aid. Would also make styling them easier, as we'd be less constrained by the behaviour of the browser default tooltip renderer, which inherits styles from the parent element and as far as I know can't be targeted directly.

Aprillion:
https://www.tpgi.com/using-the-html-title-attribute/ suggests that screen readers don't want to use `title` for accessibility due to lack of support ... I added score only for our debugging (if there are bad matches, to see if we can use some cutoff score to hide low-score matches or if score is unrelated to human evaluation).

This ticket is about replacing the built-in `title` tooltips with custom tooltips anyway, so we wouldn't be using `title` after we implement this, whether or not we keep scores.

L-Sva:
I've implemented this change, using Tippy as the new tooltip library. I've changed the root.css file to use #fed for the tooltip background colour, which looks horrendous on my computer:
![image](https://user-images.githubusercontent.com/58653425/218275294-354761ac-f120-4cc8-b26d-cf4e175b9a6e.png)

Aprillion:
@L-Sva there are 2 themes, `#fed` was intended for light theme (you can change the mode in browser dev tools to test it in light mode), feel free to use whatever doesn't look horrendous in dark theme inside the media query in CSS please

PR #153, opened 2023-02-11
Change tooltips to use Tippy instead

L-Sva:
Implements #51.

This update removes the title tag and instead displays the same information in using Tippy. This was only implemented for the search results.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "dependencies": {
     "@remix-run/cloudflare-workers": "^1.6.8",
     "@remix-run/react": "^1.6.8",
+    "@tippyjs/react": "^4.2.6",
     "copy-to-clipboard": "^3.3.2",
     "lodash": "^4.17.21",
     "markdown-it": "^13.0.1",
--- a/app/components/search.tsx
+++ b/app/components/search.tsx
@@ -4,6 +4,7 @@ import {Question} from '~/routes/questions/$question'
 import {AddQuestion} from '~/routes/questions/add'
 import {MagnifyingGlass} from '~/components/icons-generated'
 import AutoHeight from 'react-auto-height'
+import Tippy from '@tippyjs/react';
 
 type Props = {
   onSiteAnswersRef: MutableRefObject<Question[]>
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -5,6 +5,7 @@
   "dependencies": {
     "@remix-run/cloudflare-workers": "^1.6.8",
     "@remix-run/react": "^1.6.8",
+    "@tippyjs/react": "^4.2.6",
```
Aprillion, line 8: please also commit changes to `package-lock.json` after running `npm i`

ccstan99:
Maybe I missed a discussion somewhere... Could you please explain what is the goal of this enhancement? Running this on my local dev environment makes the tensorflow score result much more prominent than necessary and isn't lined up with the actual question score either. So I'm not sure this is a desirable direction. Thoughts @plexish?
Screenshot 2023-02-21 at 12 26 42 PM copy

ccstan99 (commented):
Maybe I missed a discussion somewhere... Could you please explain what is the goal of this enhancement? Running this on my local dev environment makes the tensorflow score result much more prominent than necessary and isn't lined up with the actual question score either. So I'm not sure this is a desirable direction. Thoughts @plexish?
Screenshot 2023-02-21 at 12 26 42 PM copy

plexish:
@ccstan99 this was meant to deal with #51, an issue you opened a long time ago. Might have been misunderstood though?

ccstan99:
Oh, I see. It looks like the other tooltips had already been fixed previously or at least replaced by text under icons. This PR only adds Tippy to the search results, which I think both the color & behavior unnecessarily draws attention to those scores. Nevertheless, we can close #51 anyway. Thanks!

ccstan99:
I just looked at the original issue #51  and @Aprillion & @ChrisRimmer discussing the implications on screen readers. Was Tippy supposed to resolve that? Sorry, if I closed this out prematurely. If it's only for aesthetic reasons as I originally wrote it for, then we're all set. Otherwise, this can be reopened. Thanks again!

PR #158, opened 2023-02-13
Add helpful button

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/19


.github/workflows/deploy.yml
```diff
@@ -21,7 +21,7 @@ jobs:
             | sed s/{CLOUDFLARE_ACCT_ID}/${{ secrets.CLOUDFLARE_ACCT_ID }}/ \
             | sed s/{STAMPY_KV_ID}/${{ secrets.STAMPY_KV_ID }}/ \
             | sed s/{CODA_TOKEN}/${{ secrets.CODA_TOKEN }}/ \
-            | sed s/{CODA_INCOMING_TOKEN}/${{ secrets.CODA_INCOMING_TOKEN }}/ \
+            | sed s/{CODA_WRITE_TOKEN}/${{ secrets.CODA_WRITE_TOKEN }}/ \
```
mruwnik, line 24: these will have to be changed :/ The previous tokens only allowed writing to the incoming table. An alternative would be to add a new view with just the metrics (for lack of a better word), which should be a lot safer at the cost of extra tokens

app/routes/questions/actions.tsx
```diff
@@ -0,0 +1,116 @@
...
+    title: 'Flag',
+  },
+  followup: {
+    Icon: Followup,
+    title: 'Followup',
+  },
+  helpful: {
+    Icon: Like,
+    title: 'Helpful',
+    handler: likeQuestion,
```
mruwnik, line 45: this should, in theory, make adding new actions just a matter of adding an appropriate handler function to the appropriate entry here

app/routes/questions/actions.tsx
```diff
@@ -0,0 +1,116 @@
...
+  return redirect('/')
+}
+
+export const Action = ({pageid, actionType}: {pageid: string; actionType: ActionType}) => {
+  const [remixSearchParams] = useSearchParams()
+  const [stateString] = useState(() => remixSearchParams.get('state') ?? '')
+  const {Icon, title} = actions[actionType]
+
+  // Get the state of this action for the given `pageid` from localstorage. This
+  // will result in each action only being allowed once per browser
```
mruwnik, line 81: Or would it be better to just use the basic react state?

app/routes/questions/actions.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  if (actionTaken) {
+    return <></>
+  }
+
+  return (
+    <Form replace action="/questions/actions" method="post" className="icon-link" title={title}>
+      <input type="hidden" name="action" value={actionType} />
+      <input type="hidden" name="pageid" value={pageid} />
+      <input type="hidden" name="stateString" value={stateString} />
```
mruwnik, line 109: this isn't really needed here, but I left it so users without js can also mark things as helpful

.github/workflows/deploy.yml
Aprillion: ok, you now have Admin access to be able to add more secrets to https://github.com/StampyAI/stampy-ui/settings/secrets/actions

app/routes/questions/actions.tsx
Aprillion: hm,, localstorage sounds appropriate, let's see how it works and change only if we discover some problematic edge case

Aprillion (dismissed):
if @plexish is happy that in "beta" we can release features directly to production and fix stuff as discovered, feel free to merge

if we need to start being more careful, can you set up your own free CF Worker for testing? (or let me know if there are gaps in README how to do it and I can deploy it to my test worker and try to improve documentation)

plexish:
> if @plexish is happy that in "beta" we can release features directly to production and fix stuff as discovered, feel free to merge
> 
> if we need to start being more careful, can you set up your own free CF Worker for testing? (or let me know if there are gaps in README how to do it and I can deploy it to my test worker and try to improve documentation)

Risking things being a bit broken temporarily is acceptable to not introduce overhead at this point, if there's someone around to revert if they do break.

mruwnik:
One extra thing worth mentioning - Coda doesn't seem to have transactions, so multiple writes at the same time can get overwritten. It also refreshes ~per second, so if the same question is marked helpful twice (or more) in the same ~second, it will only count as 1 helpful. Shouldn't be a problem for now (hopefully?) but worth keeping in mind.

Aprillion (dismissed):

mruwnik:
Also, the SVGs are a bit strange. This is what it looks like normally:
image

While this is what it looks like when hovering over it:
image

Should that be filled in? I could give it a go, but by no means guarantee a sanguine outcome...

Aprillion:
> Should that be filled in? I could give it a go, but by no means guarantee a sanguine outcome...

probably related to https://github.com/StampyAI/stampy-ui/commit/91b1f4820ecd67a2c81a25e5a924fa2aafcda65c#diff-a05c1079ca463546b4a858e02b8bd51c02a64c12f4ab4d573aa9d7a75451babe where we modified the svg files to contain `gray` class for correct transitioned outline in both light and dark themes

previously unused svg files still need to be manually updated - feel free to give it a try, potentially add new outlines if some icons don't have it yet

mruwnik:
How about this? 
image

Aprillion (approved):
LGTM

Issue #161, opened 2023-02-13
Drag and drop should update the URL


plexish:
This will make it easier to configure nice links to share with people

PR #163, opened 2023-02-14
Fix url bugs

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/161
https://github.com/StampyAI/stampy-ui/issues/160

### DIFF
--- a/app/hooks/stateModifiers.tsx
+++ b/app/hooks/stateModifiers.tsx
@@ -5,25 +5,28 @@ type StateString = string
 
 export const TOP = '-1'
 
-const uniqueChecker = (): ((e: StateEntry) => boolean) => {
+const makeUniqueChecker = (): ((e: StateEntry) => boolean) => {
   const seen = new Set()
   return ([pageid]) => !seen.has(pageid) && Boolean(seen.add(pageid))
 }
-export const getStateEntries = (state: StateString): StateEntry[] =>
-  Array.from(state.matchAll(/([^-_r]+)([-_r]*)/g) ?? [])
-    .map(
+export const getStateEntries = (
+  state: StateString,
+  func = (e: StateEntry[]): StateEntry[] => e
+): StateEntry[] =>
+  func(
+    Array.from(state.matchAll(/([^-_r]+)([-_r]*)/g) ?? []).map(
       (groups) =>
         [
           groups[1], // question id
           (groups[2] as QuestionState) || QuestionState.OPEN,
         ] as StateEntry
     )
-    .filter(uniqueChecker())
+  ).filter(makeUniqueChecker())
 
 export const processStateEntries = (
   state: StateString,
-  func = (e: StateEntry[]): StateEntry[] => e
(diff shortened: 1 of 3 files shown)


app/hooks/stateModifiers.tsx
```diff
@@ -5,11 +5,25 @@ type StateString = string
...
-  ])
+  Array.from(state.matchAll(/([^-_r]+)([-_r]*)/g) ?? [])
+    .map(
+      (groups) =>
+        [
+          groups[1], // question id
+          (groups[2] as QuestionState) || QuestionState.OPEN,
+        ] as StateEntry
+    )
+    .filter(uniqueChecker())
```
mruwnik, line 21: make sure that duplicates get removed from the url string. The first one found is the value (and position) in the resulting url string

app/hooks/stateModifiers.tsx
```diff
@@ -5,11 +5,25 @@ type StateString = string
 
 export const TOP = '-1'
 
+const uniqueChecker = (): ((e: StateEntry) => boolean) => {
```
Aprillion, line 8: can we call it `makeUniqueChecker` or `uniqueCheckerFactory`?

app/hooks/stateModifiers.tsx
```diff
@@ -5,11 +5,25 @@ type StateString = string
...
+          groups[1], // question id
+          (groups[2] as QuestionState) || QuestionState.OPEN,
+        ] as StateEntry
+    )
+    .filter(uniqueChecker())
+
+export const processStateEntries = (
+  state: StateString,
+  func = (e: StateEntry[]): StateEntry[] => e
+): StateString => func(getStateEntries(state)).filter(uniqueChecker()).flat().join('')
```
Aprillion, line 26: does it need double filtering, both inside `getStateEntries` and after?

Aprillion (approved):

app/hooks/stateModifiers.tsx
mruwnik: yes and no :/  which means that can be tidied up

app/hooks/stateModifiers.tsx
mruwnik: changed

app/hooks/useQuestionStateInUrl.ts
```diff
@@ -28,7 +29,9 @@ export default function useQuestionStateInUrl(minLogo: boolean, initialQuestions
   const [remixSearchParams] = useSearchParams()
   const transition = useTransition()
 
-  const [stateString, setStateString] = useState(() => remixSearchParams.get('state'))
+  const [stateString, setStateString] = useState(
+    () => remixSearchParams.get('state') && processStateEntries(remixSearchParams.get('state'))
```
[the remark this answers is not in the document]
mruwnik: it would, but would require changing the type, which would then mess up the other places...

Aprillion (approved):

PR #166, opened 2023-02-16
Add ErrorBoundary for questions

mruwnik:
Per question errors:
image

When something goes wrong on the server:
image

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -30,21 +30,18 @@ export const loader = async ({request}: Parameters<LoaderFunction>[0]) => {
   }
 }
 
-export default function App() {
-  const {minLogo} = useLoaderData<ReturnType<typeof loader>>()
-
+function Head({minLogo}: {minLogo: boolean}) {
   return (
-    <html lang="en">
-      <head>
-        <meta charSet="utf-8" />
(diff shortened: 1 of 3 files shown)


app/root.tsx
```diff
@@ -30,21 +30,18 @@ export const loader = async ({request}: Parameters<LoaderFunction>[0]) => {
...
-          dangerouslySetInnerHTML={{
-            __html: `
+    <head>
+      <meta charSet="utf-8" />
+      <meta name="viewport" content="width=device-width,initial-scale=1" />
+      {/* https://github.com/darkreader/darkreader/issues/1285#issuecomment-761893024 */}
+      <meta name="color-scheme" content="light dark" />
+      <Meta />
+      <Links />
+      <script
```
mruwnik, line 42: why is this here, rather than at the bottom of the page? It would be nice to move it down, to get rid of the error that appears with each page reload

Aprillion: I was not involved in adding HotJar myself, but I found https://help.hotjar.com/hc/en-us/articles/115009336727-How-to-Install-Your-Hotjar-Tracking-Code recommends to add it in the `<head>` ... no explanation why

Aprillion (approved):

Issue #165, opened 2023-02-16
"Edit" link doesn't open Google Docs


Mihonarium:
Clicking on the "EDIT" link (not the pencil icon but the text) moves the page to the start and doesn't open Google Docs.

plexish:
aha, that makes sense

PR #167, opened 2023-02-16
Allow edit links to be clicked

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/165

### DIFF
--- a/app/routes/index.tsx
+++ b/app/routes/index.tsx
@@ -110,7 +110,7 @@ export default function App() {
 
   const handleSpecialLinks = (e: MouseEvent) => {
     const el = e.target as HTMLAnchorElement
-    if (el.tagName !== 'A') return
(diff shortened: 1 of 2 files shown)


app/routes/index.tsx
```diff
@@ -110,7 +110,7 @@ export default function App() {
 
   const handleSpecialLinks = (e: MouseEvent) => {
     const el = e.target as HTMLAnchorElement
-    if (el.tagName !== 'A') return
+    if (el.tagName !== 'A' || el.closest('.question-footer')) return
```
mruwnik, line 113: Dunno about this. I was thinking of introducing some kind of constant or something, in case the class name changes

Aprillion: probably best if the <a> element has some special class used only for this purpose.. but whatever works is fine I guess

Aprillion (approved):

Issue #164, opened 2023-02-16
Gracefully handle unknown questions


mruwnik:
If someone manually inputs the id of an non existent question (e.g. https://aisafety.info/?state=ZZZZ), the UI will keep showing that it's loading:

<img width="650" alt="image" src="https://user-images.githubusercontent.com/3942390/219382033-f402d7b2-0fd5-4972-b143-6cf224feb0bb.png">

# Suggested fixes:
* ignore it and remove it from the URL - this is iffy, as it's magical and looks like a bug
* ignore it and just don't display an entry for that page id - also iffy, for the same reason
* display some kind of error message - this is nice in so far as it's explicit, but can result in clutter

plexish:
Agree with the analysis, I think I lean towards option 3. Something like _No question found with ID $1. Please go to the Discord in the lower right and report where you found this link._

PR #168, opened 2023-02-16
handle unknown pageids

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/164

image

### DIFF
--- a/app/routes/questions/$question.tsx
+++ b/app/routes/questions/$question.tsx
@@ -20,6 +20,7 @@ export const loader = async ({request, params}: Parameters<LoaderFunction>[0]) =
     return await loadQuestionDetail(request, question)
   } catch (error) {
     return {
+      error: error.toString(),
       timestamp: new Date().toISOString(),
       data: {
         pageid: question,
@@ -36,7 +37,10 @@ export const loader = async ({request, params}: Parameters<LoaderFunction>[0]) =
 export function fetchQuestion(pageid: string) {
   const url = `/questions/${encodeURIComponent(pageid)}`
   return fetch(url).then(async (response) => {
-    const {data, timestamp}: Awaited<ReturnType<typeof loader>> = await response.json()
+    const {data, timestamp, error}: Awaited<ReturnType<typeof loader>> = await response.json()
+
+    if (error) console.error(error)
+
     reloadInBackgroundIfNeeded(url, timestamp)
 
     return data
--- a/app/routes/questions/$question.tsx
+++ b/app/routes/questions/$question.tsx
@@ -1,5 +1,5 @@
 import type {LoaderFunction} from '@remix-run/cloudflare'
-import {loadQuestionDetail} from '~/server-utils/stampy'
+import {loadQuestionDetail, QuestionStatus} from '~/server-utils/stampy'
 import {useRef, useEffect, useState} from 'react'
 import AutoHeight from 'react-auto-height'
 import type {Question} from '~/server-utils/stampy'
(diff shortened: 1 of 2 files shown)


app/routes/questions/$question.tsx
```diff
@@ -16,7 +16,21 @@ export const loader = async ({request, params}: Parameters<LoaderFunction>[0]) =
     throw Error('missing question title')
   }
 
-  return await loadQuestionDetail(request, question)
+  try {
+    return await loadQuestionDetail(request, question)
+  } catch (error) {
+    return {
+      timestamp: new Date().toISOString(),
```
mruwnik, line 23: I would have preferred to do this later, e.g. [here](https://github.com/StampyAI/stampy-ui/pull/168/files#diff-1f32e8eae48dea5606d6c249c24da8eff4f4d90c2e49dd6764e8cd417f99059cR117), but it complicated things a lot, so I just threw it in here

app/routes/questions/$question.tsx
```diff
@@ -110,8 +124,8 @@ export function Question({
   }
 
   return (
-    <article className={cls} {...props}>
-      <h2 onClick={handleToggle} title={isExpanded ? 'Hide answer' : 'Show answer'}>
+    <article className={cls}>
+      <h2 onClick={handleToggle} title={isExpanded ? 'Hide answer' : 'Show answer'} {...dragProps}>
```
mruwnik, line 128: currently the whole answer is draggable, which in turn means that you can't select the contents to e.g. copy a paragraph. This will make only the header draggable

app/routes/questions/$question.tsx
```diff
@@ -134,26 +148,28 @@ export function Question({
...
-                      href={answerEditLink}
-                      target="_blank"
-                      rel="noreferrer"
-                      title="edit answer"
-                    >
-                      <Edit />
-                      Edit
-                    </a>
-                  )}
+              {text !== null && questionProps.status != QuestionStatus.UNKNOWN && (
```
mruwnik, line 151: no point in having actions for questions that aren't actionable

app/routes/questions/$question.tsx
Aprillion: we also need to send the error to the browser, e.g. in additional property `error`, it's much easier to inspect network response than to inspect server logs of cloudflare workers

Aprillion (approved):

app/routes/questions/$question.tsx
```diff
@@ -16,13 +16,31 @@ export const loader = async ({request, params}: Parameters<LoaderFunction>[0]) =
     throw Error('missing question title')
   }
 
-  return await loadQuestionDetail(request, question)
+  try {
+    return await loadQuestionDetail(request, question)
+  } catch (error) {
+    return {
+      error: error.toString(),
```
mruwnik, line 23: this is a slight potential security issue, if there are sensitive stuff in the error message. Should be fine?

app/routes/questions/$question.tsx
mruwnik: added

app/routes/questions/$question.tsx
Aprillion: We only forward reponses from Coda, not the request with keys, and if someone wanted to send a bunch of spam likes, they can misuse the current system without generating any error messages, so I don't believe forwarding errors is a significant risk at this point... We might need a security audit if Stampy becomes more popular though, the feature to allow anonymous access might not survive if we start dealing with trolls or similar.

PR #170, opened 2023-02-18
Fix TypeScript errors

Aprillion:


mruwnik (dismissed):

mruwnik (approved):

Aprillion:
@mruwnik aah, most APIs don't have `Status` field in the response, so I'm removing assigning the `Unknown` value by default, keeping it undefined, so that we display action buttons

feel free to investigate in Coda, if more of the APIs need that column added

mruwnik:
this is why I dislike strict static typing - it tends to be over restrictive :D

mruwnik (approved):

PR #172, opened 2023-02-19
Allow retraction of marked helpful answers

mruwnik:


app/routes/questions/actions.tsx
```diff
@@ -92,22 +93,27 @@ export const Action = ({pageid, actionType}: {pageid: string; actionType: Action
...
 
-    setActionTaken(true)
-    const searchParams = new URLSearchParams({pageid, action: actionType})
+    const incBy = actionTaken ? '-1' : '1'
+    setActionTaken(!actionTaken)
+
+    // This sort of cheats - if more than 1 request is sent per second (or some other such time
+    // period), one of them will be (sort of) picked at random. This should be ok in the long run.
+    // Hopefully.
+    const searchParams = new URLSearchParams({pageid, incBy, action: actionType})
```
Aprillion, line 102: I hope the backend is not blindly accepting any number anyone can send via API a la `-10000`, right?!?

Aprillion (approved):

app/routes/questions/actions.tsx
mruwnik: what? Of course it is :D How much would you limit it to? That being said, there's nothing stopping someone from writing a script to just send -1 in a loop :/

Aprillion: if it was up to me, I would allow only `-1` and `1` as enum options (or rather `action: inc` and `action: dec`) and only 1 operation per CSRF token that would have to be fetched from Coda with some warnings sent to Discord if a soft limit of operations per second is reached and self-disabling feature if some hard limit is detected 😓

Issue #171, opened 2023-02-18
Answer text sometimes fails to load, and sometimes unloads after loading


ChrisRimmer:
Initially reported by Plex in Discord: https://discord.com/channels/677546901339504640/758062805810282526/1076611649487786085

## Steps to reproduce:

1. Clear cache at https://ui.stampy.ai/questions/cache
2. Load https://aisafety.info/?state=2222_
3. If the question loaded correctly and the whole answer text appears, return to step 1

## Expected behaviour:
- The answer text loads reliably

## Actual behaviour:
- Occasionally the answer text fails to load at all (insufficient testing to provide a %age rate)
- (Very?) rarely, the answer text loads, but then "unloads" and is replaced with the placeholder "Loading..." string

## Tests performed and other notes:
- API response from https://aisafety.info/questions/2222 seems to be correct even when the answer text doesn't render
- Tested on at least three devices (Chris' iPhone and desktop, running Safari and Firefox respectively) plus whatever Plex uses

I suspect the steps to reproduce above are incomplete, as the failure rate seemed to slowly go to zero as I repeated the process. In my last ten or so attempt to clear my question cache and reload the page, the answer has loaded correctly every time, after failing quite reliably earlier on. I think the error smells a bit like a race condition, so if there's some network or application level caching happening that might have affected the timing of the network request I can imagine that might be responsible for the increase in my perceived reliability. Just a guess though!

Aprillion:
It could be the way how infinite scroll de-duplicates answers, since the responses have `text: null` e.g.:

e.g. https://aisafety.info/questions/allOnSite?nextPage=https://coda.io/apis/v1/docs/fau7sl2hmG/tables/table-1Q8_MjxUes/rows?pageToken=eyJsaW1pdCI6MTAsIm9mZnNldCI6MTAsIm9wVmVyc2lvbiI6MzMxMjIsInNjaGVtYVZlcnNpb24iOjE3MSwic29ydEJ5IjoibmF0dXJhbCIsInVzZUNvbHVtbk5hbWVzIjp0cnVlLCJ2YWx1ZUZvcm1hdCI6InJpY2gifQ 
![image](https://user-images.githubusercontent.com/1087670/219939611-fbd4be5a-3d02-4bc1-bc78-a10361812b69.png)

@mruwnik can you please investigate?

mruwnik:
Hopefully fixed by the above merge - if this bug doesn't reappear over the next week or so, this can be closed

[...]

PR #173, opened 2023-02-19
Fix disappearing items

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/171

- tags
- eternal "loading..."
- fetch tags per show


app/hooks/useQuestionStateInUrl.ts
```diff
@@ -91,9 +90,9 @@ export default function useQuestionStateInUrl(minLogo: boolean, initialQuestions
       text: null,
       answerEditLink: null,
       relatedQuestions: [],
-      ...questionMap.get(pageid),
       questionState,
       tags: [],
+      ...questionMap.get(pageid),
```
mruwnik, line 95: this was overwriting the question's tags, resulting in them sometimes being shown, sometimes not

app/routes/tags/$tag.tsx
```diff
@@ -0,0 +1,124 @@
...
+        Select a question from the <b>{tag.name}</b> tag...
+      </div>
+      {children}
+    </Dialog>
+  )
+
+  if (!tag.rowId) {
+    return (
+      <TagDialog>
+        <div className="loader"></div>
```
mruwnik, line 65: display a loader thingy until the tag's questions have been fetched

app/server-utils/stampy.ts
```diff
@@ -256,14 +255,15 @@ const toTag = (r: CodaRow, nameToId: Record<string, string>): Tag => ({
     .filter((q) => q.pageid),
   mainQuestion: extractMainQuestion(r.values['Main question']),
 })
-export const loadAllTags = withCache('allTags', async () => {
```
mruwnik, line 259: the tags -> questions used to work by first fetching all tags and questions, then making a big, global map for them. This will now fetch each tag separately once it gets clicked on. It's slower (at least the first time a tag is displayed), but is less magical

app/server-utils/stampy.ts
```diff
@@ -94,15 +94,14 @@ const CODA_DOC_ID = 'fau7sl2hmG'
 // Use table ide, rather than names, in case of renames
 const QUESTION_DETAILS_TABLE = 'grid-sync-1059-File' // Answers
 const INITIAL_QUESTIONS_TABLE = 'table-yWog6qRzV4' // Initial questions
-const ON_SITE_TABLE = 'table-1Q8_MjxUes' // On-site answers
+const ON_SITE_TABLE = 'table-aOTSHIz_mN' // On-site answers
```
mruwnik, line 97: I'm not sure, but I think this was causing the hanging "loading..." - the "On site" table doesn't have the text nor tags - the "On site API" table should be used, which has the same filters applied, but also returns the text

Aprillion: huh,, this sounds like it might solve the race condition, let's see if people will still be able to randomly reproduce afterwards (hopefully not 🍀)

app/routes/tags/$tag.tsx
Aprillion: if I understand correctly, we always display question tags, but then lazy load the other questions for a given tag if you click the tag?

that sounds like correct behaviour to me, no unnecessary loading for people not interested in tags, but no loader when opening just the question itself 👍

Aprillion (approved):

app/routes/tags/$tag.tsx
mruwnik: exactly. It looks like this:
image

Issue #79, opened 2022-08-02
Switch to using the [See more...] expansion


plexish:
Parser should capture the text `[See more...]` (with or without dots and with any capitalization, e.g. `[see more]` should work), and make a javascript element which on click displays the remainder of the answer. There was an element which did this previously with the wiki backend, which might be useful.

Ideally has the URL change functionality described in #69 and no longer has the bug described in #68

PR #177, opened 2023-02-20
Hide everything after [See more...]

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/79

image

### DIFF
--- a/app/routes/index.tsx
+++ b/app/routes/index.tsx
@@ -106,6 +106,15 @@ export default function App() {
     .map(({title}) => title)
 
   const handleSpecialLinks = (e: MouseEvent) => {
+    // The AutoHeight component doesn't notice when a HTML details is opened.
+    // Manually removing the height from the style fixes this, but can potentially
+    // break something else...
+    const target = e.nativeEvent.target as HTMLElement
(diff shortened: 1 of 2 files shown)


app/routes/index.tsx
```diff
@@ -106,6 +106,15 @@ export default function App() {
     .map(({title}) => title)
 
   const handleSpecialLinks = (e: MouseEvent) => {
+    // The AutoHeight component doesn't notice when a HTML details is opened.
+    // Manually removing the height from the style fixes this, but can potentially
+    // break something else...
+    const target = e.nativeEvent.target as HTMLElement
```
mruwnik, line 112: This I'm very unsure about... But it works, at least for now...

Aprillion: if it works for now, let's try it and figure out problems as they come, fortunately it's not AGI Alignment

Aprillion (approved):

plexish:
@mruwnik could the text be blue [See more...], and vanish once you click it? I think that'd feel like a more standard interface. It's OK that it can't be re-hidden without refresh (or later modifying URL)

Issue #38, opened 2022-06-30
+1 for unanswered canonical questions in search


plexish:
Increment `Request Count` in the coda answers table in response to this being pressed. It's OK for readers to request multiple times, as we don't have session data.

PR #179, opened 2023-02-20
Show additional, unanswered questions in search

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/38

Search bar:

image

Loading screen:

image

Choose an alternative:

image


app/components/search.tsx
```diff
@@ -171,13 +194,86 @@ const ResultItem = ({
...
+  onClose,
+  onSelect,
+}: {
+  question: string
+  relatedQuestions: string[]
+  openQuestionTitles: Props['openQuestionTitles']
+  onSelect: Props['onSelect']
+  onClose: (e: any) => void
+}) => {
+  const title = `Does any of the following questions match your question?`
```
mruwnik, line 217: @plexish @ccstan99 what should this be?

app/components/search.tsx
```diff
@@ -171,13 +194,86 @@ const ResultItem = ({
...
+  const [extraQuestions, setExtraQuestions] = useState<SearchResult[]>(empty)
+
+  useEffect(() => {
+    const getResults = async (question: string) => {
+      try {
+        const results = await (
+          await fetch(`/questions/search?question=${encodeURIComponent(question)}`)
```
mruwnik, line 224: I wanted to go directly to NLP, but CORS...

app/components/search.tsx
```diff
@@ -171,13 +194,86 @@ const ResultItem = ({
...
+        ).json()
+        setExtraQuestions(results)
+      } catch (error) {
+        console.error(error)
+        setExtraQuestions([])
```
mruwnik, line 229: what should happen if the request fails? This will result in an empty list, but I'm guessing an error should be displayed?

app/routes/questions/actions.tsx
```diff
@@ -39,10 +40,15 @@ const actions = {
     Icon: Followup,
     title: 'Followup',
   },
+  ping: {
```
mruwnik, line 43: this could do with a better name...

app/routes/questions/actions.tsx
```diff
@@ -39,10 +40,15 @@ const actions = {
...
+    Icon: Like,
```
mruwnik, line 44: which icon should be used? There isn't a +1 icon, so I just used the thumbs up one for now

app/routes/questions/search.tsx
```diff
@@ -0,0 +1,19 @@
...
+  const url = new URL(request.url)
+  const question = url.searchParams.get('question')
+
+  if (!question) return []
+
+  return await search(question)
+}
+
+export function search(question: string) {
+  const url = `${NLP_SEARCH_ENDPOINT}/api/search?query=${question}`
```
mruwnik: @ccstan99 is this the right endpoint to use for this?

app/components/search.tsx
```diff
@@ -93,54 +97,73 @@ export default function Search({onSiteAnswersRef, openQuestionTitles, onSelect}:
...
+                  />
+                ))}
+            </div>
+            <button
+              className="result-item none-of-the-above"
+              onClick={() => setShowMore(true)}
+              onMouseDown={() => setHide(false)}
+              onMouseUp={() => setHide(true)}
+            >
+              Show more answers
```
mruwnik, line 149: @plexish @ccstan99 This is displayed under the results for stuff that are live on the site - I'm hoping you'll come up with a better text

app/components/search.tsx
ccstan99: How about "Here are some questions we're still answering. Do any of these match what your question?"

app/routes/questions/actions.tsx
ccstan99: upvote?

app/components/search.tsx
ccstan99: The NLP endpoint will always try to return the 10 closest questions. If it fails, we can either (1) return an error or preferably (2) add the question to suggestions, let the user know and thank them.

app/routes/questions/search.tsx
ccstan99: Yes, but we should probably filter out already "Live on Site" questions shown in the previous step. I still need to add that parameter option for you.

app/components/search.tsx
ccstan99: Maybe "More options..."

app/routes/questions/actions.tsx
ccstan99: OK, that's probably fine for now. I can create something different later if we decide that's needed. However, that icon should also to be selected/highlighted when the cursor is hovering over the questions, because clicking the question and icons performs the same function, correct?

ccstan99:
Nice! We should also include the user's original question at the top of the modal dialogue, just for their reference.

app/routes/questions/actions.tsx
plexish: Request?

app/components/search.tsx
plexish: hmm.. 
> I'm asking something else

Maybe?

app/components/search.tsx
plexish: or maybe
> Here are some questions we're still answering. Are any of these what you're asking?

to avoid the double use of "question"

mruwnik:
how about something like this:

image

app/components/search.tsx
mruwnik: I went with option 2

app/routes/questions/actions.tsx
mruwnik: changed to request

app/routes/questions/search.tsx
mruwnik: let me know when it's ready. Currently it just manually filters them out, which will result in various lengths of displayed questions

app/components/search.tsx
mruwnik: changed to "I'm asking something else"

app/components/search.tsx
```diff
@@ -171,13 +194,95 @@ const ResultItem = ({
...
+      </Dialog>
+    )
+  }
+  return (
+    <Dialog onClose={onClose}>
+      <div className="dialog-title">
+        You searched for &quot;{question}&quot;.
+        <br />
+        Here are some questions we&apos;re still answering. Are any of these what you&apos;re
+        looking for?
```
mruwnik, line 253: @plexish @ccstan99 is this better?

ccstan99:
Looks good to me. Should we include a "Cancel" or "Nevermind" too? Maybe to the right of "+ None of these..."? Or an "x" in the upper right? Or does clicking outside the dialogue close it?

mruwnik:
clicking outside will close it. But that might not be intuitive...

app/routes/questions/search.tsx
ccstan99: Unfortunately, I probably won't get to this until the weekend. But I'll definitely let you know!

ccstan99:
I'm ok with that for now, as long as they have an option to leave and aren't forced too request a new question.

ccstan99:
Just ran this on my local dev. It looks like clicking on the question in the 1st column actually displays it in the UI browser window, even if it's unanswered (which says "Loading...") or messy (if in-progress). @plexish Do we really want non-canonically answered questions displaying in the UI browser? I thought both clicking the question & request would do the same thing or did you decide differently?

mruwnik:
no, I just forgot about it :/ On it now

ccstan99:
Great, thanks! It would be good to have hovering over the question AND the request icon to get highlighted together to make it clear they do the same thing.

mruwnik:
Took longer than I expected...

image

plexish:
>  "in the upcoming weeks" instead?
Yeah, that is probably better.

mruwnik:
image

image

image

ccstan99 (approved):
Thanks for taking care of this! 
Couple minor things that can be addressed later as a separate issue:
1. Vertical height of dialogue shrinks after switching to thank you message, would be nice to also stay fixed.
2. There probably should be a more explicit way (x in upper right and/or OK/Close button) to close the dialogue after the question has been requested/suggested. Not too critical since clicking outside closes it anyway.

Issue #180, opened 2023-02-21
Styling see more


plexish:
@mruwnik could the text be blue [See more...], and vanish once you click it? I think that'd feel like a more standard interface, and also be cleaner for people copying text out of the UI. It's OK that it can't be re-hidden without refresh (or later modifying URL)

Issue #69, opened 2022-07-11
URL change for [See more...], so you can link directly to detailed answer


plexish:
I noticed myself wanting to do this as a reply to this bounty: https://www.facebook.com/groups/bountiedrationality/posts/3247277178849656/, with the short version of https://ui.stampy.ai/?state=6703_ being insufficient. Might be mostly a sign to only have a long answer to that particular question for now.

Maybe a + in the URL for detailed?

Aprillion:
I don't like this, can we avoid short/long details in the first place, then we won't need this feature either?

plexish:
hm, I have intuitions that as we get a bunch of editors we'll wish we had this split as the answers gradually get longer with valuable but non-essential information, and that some readers will want far more detail than others.. but I'm happy to put development on this on ice until we end up getting requests in that direction as my intuitions might be wrong.

Would you prefer to use just the briefest description on the UI and remove the existing toggle for now,?

ccstan99:
I can totally understand why some editors and readers will want more details. However, from a UX perspective, it feels odd to have a very long detailed answer embedded into a list with other collapsable Q&A. Given our current setup, it'd be preferable to have long answers chunked to smaller related Q&A. For example, the bulk of the long answer to "How does AI taking things literally contribute to alignment being hard?" is already in the link to [Goodhart's law](https://stampy.ai/read/Goodhart%27s_law). BTW, will all the content in https://stampy.ai/read remain or is the new UI supposed to replace it?

Alternatively, how would you guys feel about displaying maybe the 1st 2000 (or some n, less for mobiles) characters followed by ... link to the detailed answer on its own page either in the wiki or wherever? That's a common convention for news and blog indices. We already have the precedence of people leaving the UI to follow other links within answers. This might encourage editors to be more concise (within 2000 or some n chars) but still allows us to serve answers that aren't brief enough without overwhelming the UI. 

Thoughts?

plexish:
Agreed that we want to encourage editors to chunk things where possible.

> how would you guys feel about displaying maybe the 1st 2000 (or some n, less for mobiles) characters followed by ... link to the detailed answer on its own page either in the wiki or wherever?

oo, I like this idea. ... is more well-known as a UI than [Show detailed], for sure. Can we make it expand in the page itself though, rather than taking them to a new page? And do we want different cutoff points for mobile? This will make it harder for editors to design a good experience for all readers, though I could put a marker at both of these points in the wiki.

[...]

PR #181, opened 2023-02-21
Format [Show more...] as link

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/180

image

to this

image

### DIFF
--- a/app/root.css
+++ b/app/root.css
@@ -545,6 +545,11 @@ a:focus-visible,
   border-radius: 2px;
 }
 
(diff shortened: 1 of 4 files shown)


plexish:
> @plexish like this? One thing to note is that this will be reset each time the question gets collapsed

Yup, looks good. Resetting on refresh is fine for now, though with #69 it would be reflected in the URL so be persistant.

Aprillion (approved):

PR #184, opened 2023-02-23
md-to-html refactor

jrhender:
IMHO, `stampy.ts` is getting a tad long, so I had the idea to split/group the md-to-html bits into a separate directory. My plan is to have **no** functional changes with this PR.

I'm also thinking of refactoring `wrapInDetails` and `urlToIframe` into decorators but I'm not sure how that will look
https://refactoring.guru/design-patterns/decorator/typescript/example
My intuition is that `wrapInDetails` and `urlToIframe` are similar operations that are enhancing the html in some way so it would be good to organize the code to represent this.

Part of my rational for this refactoring is also that having the app code a bit more organized might make it easer to organize tests for the code 🙈 

Please let me know if any thoughts on this, just in the ideation phase at this point.


Aprillion:
in principle, whatever refactoring you can agree with @mruwnik is fine by me

however, it might be better to start adding tests before refactoring, otherwise not sure how confident we can be everything will work the same...

jrhender:
@Aprillion great point re having tests first 👍 . I will open an issue in `stampy-ui` to discuss tests around the functionality in `stampy.ts`

jrhender:
I will close this PR as it should probably be done with tests in place and it might take a little while to get those.

PR #185, opened 2023-02-25
Loading for search

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/183

When nothing loaded:
image

When questions present:
image

### DIFF
--- a/app/server-utils/url-to-iframe.ts
+++ b/app/server-utils/url-to-iframe.ts
@@ -1,5 +1,5 @@
 /**
- * Replaces all of the URLs in text with iframes of the URLs
+ * Replaces all whitelisted URLs in the text with iframes of the URLs
  * @param text text in which URLs should be replaced
  */
 export const urlToIframe = (text: string): string => {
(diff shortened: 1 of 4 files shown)


app/server-utils/url-to-iframe.ts
```diff
@@ -1,29 +1,34 @@
...
+  //   (              <-- capture the url
+  //     https?:\/\/  <-- check for both http and https
+  //     [\.\w]+?     <-- any string of alphanumerical characters and full stops
+  //     \/?[^\]]*    <-- optional query params
+  //   )
+  // \]
+  // \(
+  //     \1           <-- lookback to match whatever the first group caught
+  // \)
+  const linkRegex = new RegExp(/\[(https?:\/\/[.\w]+?\/?[^]]*?)\]\(\1\)/gi)
```
mruwnik, line 21: this wasn't working for me, so I refactored it to work on markdown

Aprillion: if using regex literal `/.../gi`, the `new RegExp(string)` wrapper is not needed - please choose one of the syntaxes, not both

Aprillion (approved):

app/server-utils/url-to-iframe.ts
```diff
@@ -1,29 +1,33 @@
 /**
- * Replaces all of the URLs in text with iframes of the URLs
+ * Replaces all whitelisted URLs in the text with iframes of the URLs
```
Aprillion, line 2: (minor) "allowed URLs" would use less jargon for the same explanation

app/server-utils/url-to-iframe.ts
```diff
@@ -1,29 +1,33 @@
...
+  //   (              <-- capture the url
+  //     https?:\/\/  <-- check for both http and https
+  //     [\.\w]+?     <-- any string of alphanumerical characters and full stops
+  //     \/?[^\]]*    <-- optional query params
+  //   )
+  // \]
+  // \(
+  //     \1           <-- lookback to match whatever the first group caught
+  // \)
+  const matches = text.matchAll(/\[(https?:\/\/[.\w]+?\/?[^]]*?)\]\(\1\)/gi)
```
Aprillion, line 21: do we want to replace only links in the `[url](url)` format on purpose, keeping `[text](url)` as text links?

Aprillion (approved):

app/server-utils/url-to-iframe.ts
mruwnik: yes. The idea is that there are a whitelisted collection of urls that are to be transformed into iframes, but only if both the url and text are the same. So you could do `[here's a map](http://aisafety.world/)` in the main text, without it then being transformed into an iframe of the map

Issue #175, opened 2023-02-19
Searches have a fair amount of lag


plexish:
Seems considerably more than it used to be a few weeks ago? @ccstan99 has anything relevant changed?

Aprillion:
I removed the initial plain-text search because people complained about the transition (the suggestions changed after a second for initial search)... Do you perceive the problem for all searches or only for the initialization?

If only initialization, we could add a loading indicator I guess,, if all searches then not sure, but in case it's because of using larger (more accurate) encodings/model/something, and we won't find easy solution, and if we are more confident in the API solution now, then it might be a good opportunity to try switching to API solution and compare the timings.

ccstan99:
I haven't changed anything recently. There's been a recent release of tfjs (TensorFlow), so it's possible we should pin to a previous version in case something new is causing an issue. Otherwise, the model used shouldn't be larger or more accurate than before. The search space with the number of questions has grown but not signifcantly so. 

If it's only a lag in initialization, we might want to put the plain-text search back? Having an unexpected update seems minor compare to not being able to search at all. 

If it's a general lag, we could try with the API and compare. I was hoping to try testing that out later next week, if I have some time. The API does use a larger, supposedly more accurate model, but I don't expect it to be faster than running TF in-browser. Again, the main concern was all the traffic with live-updating as the user types, but certainly worth exploring.

plexish:
Just initial search, but having had the tab open for a while still causes a long wait, is the initial load only happening once a person starts typing?

I think keyword to start with and accepting there will be one update is overall better, but not using keyword at all once the initial load is complete, might be best?

tayler6000:
I noticed this as well, and it kinda gave me the impression Stampy didn't know anything.  I think a loading indicator would be a good addition to help prevent this conception to new comers.

Issue #67, opened 2022-07-10
Search results updating multiple times without extra typing


plexish:
e.g. if you search:

how do i help

and wait long enough, it updated three times re-sorting the results one time, and twice on two other occasions. Updating once after typing ends is probably unavoidable, but more than once seems odd?

ChrisRimmer:
Doing some investigation on this and while I can only get it to update the results list once after it opens, that still seems like one time too many.

So far I've got as far as working out that the first thing we see (just for a moment) is the list of results according to a plain text search, which quickly gets replaced by the tensorflow search result. I think we should defer displaying plain text results until tensorflow fails (which in fairness, I don't expect it would do often), but I don't know how tf error handling works. @Aprillion you got any ideas here?

Aprillion:
The point of plain text search is to display "something" in the first few seconds while tensorflow was loading the ~30 MB model + in case something goes wrong and the external google model never loads.

I am proposing to delete the tensorflow - temporarily in #124 while we are discussing how to implement search in Google Docs architecture, but I am a proponent of BE search without client-side tensorflow ;) We still need to load a list of questions for other features down the line, so I wouldn't remove plain-text search yet, but integration of the 2 searches in more seamless way (or completely dropping the client-side plain-text search too) should wait until we finalize the new search for GDocs.

plexish:
I just tested again and got two updates.

ccstan99:
The debounce was supposed to help with this, but it was also causing the lag issues from #175. I dropped it down to 100ms from 500ms.

I also added the original userQuery along with the TF results. Maybe we can also keep a timestamp of when the results were last updated? Ideally, when receiving results from the tfWorker, we should compare the userQuery with current search input value and only display results if it hadn't been updated too recently, i.e. the timestamp was longer than some specified (100ms or 300ms?) ago.

My life & work outside of Stampy is ramping up, so I'm not as responsive as I wish to be and am afraid I might be blocking progress on this. If someone (@mruwnik?) with more availability can resolve the rest of this issue, that'd be super helpful.

PR #186, opened 2023-02-26
warm start search, add showLive=0

ccstan99:
1. After loading, warm up model with a dummy search to speed up later searches
2. Experimenting with debounce values, was changed from 400 to 500 on 1/22
3. Add showLive=0 for semantic search endpoint so suggesting new question only show non-canonically answered questions where status is NOT "Live on site"

### DIFF
--- a/app/components/search.tsx
+++ b/app/components/search.tsx
@@ -24,9 +24,9 @@ type SearchResult = Question & {
 }
 
 type WorkerMessage = {
-  searchResults: {title: string; pageid: string; score: number}[]
+  searchResults?: {title: string; pageid: string; score: number}[]
   status: string
-  numQs: number
+  numQs?: number
 }
 
 const empty: [] = []
--- a/app/components/search.tsx
+++ b/app/components/search.tsx
@@ -23,12 +23,11 @@ type SearchResult = Question & {
   url?: string
 }
 
-type WorkerMessage =
-  {
-    searchResults: {title: string; pageid: string; score: number}[]
-    status: string
-    numQs: number
-  }
+type WorkerMessage = {
+  searchResults: {title: string; pageid: string; score: number}[]
+  status: string
+  numQs: number
+}
 
 const empty: [] = []
 
--- a/app/components/search.tsx
+++ b/app/components/search.tsx
@@ -80,7 +80,7 @@ export default function Search({onSiteAnswersRef, openQuestionTitles, onSelect}:
   }
 
   const handleChange = debounce(searchFn, 100)
-  
+
   const results = tfFinishedLoadingRef.current ? searchResults : baselineSearchResults
   const model = tfFinishedLoadingRef.current ? 'tensorflow' : 'plaintext'
 
--- a/app/components/search.tsx
+++ b/app/components/search.tsx
@@ -23,11 +23,16 @@ type SearchResult = Question & {
   url?: string
 }
 
-type WorkerMessage = {
-  searchResults?: {title: string; pageid: string; score: number}[]
-  status: string
-  numQs?: number
-}
+// type of postMessage() values from tfWorker.js (manually synchronized)
+export type WorkerMessage =
+  | {
+      status: 'ready'
+      numQs: number
+    }
+  | {
+      searchResults: {title: string; pageid: string; score: number}[]
+      userQuery?: string
+    }
 
 const empty: [] = []
 
@@ -44,8 +49,8 @@ export default function Search({onSiteAnswersRef, openQuestionTitles, onSelect}:
     const handleWorker = (event: MessageEvent<WorkerMessage>) => {
       const {data} = event
       console.debug('onmessage from tfWorker:', data)
-      if (data.status === 'ready') {
-        tfFinishedLoadingRef.current = true
+      if ('status' in data) {
+        tfFinishedLoadingRef.current = data.status === 'ready'
         return
       }
       if (data.searchResults) {
--- a/app/components/search.tsx
+++ b/app/components/search.tsx
@@ -79,7 +79,8 @@ export default function Search({onSiteAnswersRef, openQuestionTitles, onSelect}:
     }
   }
 
-  const handleChange = debounce(searchFn, 500)
+  // const handleChange = debounce(searchFn, 300)
+  const handleChange = searchFn
 
   const results = tfFinishedLoadingRef.current ? searchResults : baselineSearchResults
   const model = tfFinishedLoadingRef.current ? 'tensorflow' : 'plaintext'
--- a/app/components/search.tsx
+++ b/app/components/search.tsx
@@ -45,7 +45,7 @@ export default function Search({onSiteAnswersRef, openQuestionTitles, onSelect}:
     const handleWorker = (event: MessageEvent<WorkerMessage>) => {
       const {data} = event
       console.debug('onmessage from tfWorker:', data)
-      if (data === 'ready') {
+      if (data.status === 'ready') {
         tfFinishedLoadingRef.current = true
         return
       }
@@ -79,9 +79,8 @@ export default function Search({onSiteAnswersRef, openQuestionTitles, onSelect}:
     }
   }
 
-  // const handleChange = debounce(searchFn, 300)
-  const handleChange = searchFn
-
+  const handleChange = debounce(searchFn, 100)
+  
   const results = tfFinishedLoadingRef.current ? searchResults : baselineSearchResults
   const model = tfFinishedLoadingRef.current ? 'tensorflow' : 'plaintext'
 
--- a/app/components/search.tsx
+++ b/app/components/search.tsx
@@ -24,11 +24,11 @@ type SearchResult = Question & {
 }
 
 type WorkerMessage =
-  | 'ready'
-  | {
-      searchResults: {title: string; pageid: string; score: number}[]
-      numQs: number
-    }
+  {
+    searchResults: {title: string; pageid: string; score: number}[]
+    status: string
+    numQs: number
+  }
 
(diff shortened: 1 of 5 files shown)


app/components/search.tsx
```diff
@@ -79,7 +79,8 @@ export default function Search({onSiteAnswersRef, openQuestionTitles, onSelect}:
     }
   }
 
-  const handleChange = debounce(searchFn, 500)
+  // const handleChange = debounce(searchFn, 300)
+  const handleChange = searchFn
```
Aprillion, line 83: have you tested in Firefox and on mobile browsers? debounce was added because search was still running for out-of-date queries while typing additional letters and then it took up to 2 seconds of blinking results until the final results were shown

to search without delay, you can always press Enter

not sure why I changed it to 500ms at the same time then adding the Enter trigger,, probably left over by mistake while testing - feel free to use 300 or even 100ms, but I am not sure about the zero debounce

tayler6000:
@ccstan99 @Aprillion Would it be fair to say this PR should be linked to #175 ?

app/components/search.tsx
ccstan99: Sorry, I had this PR as a draft because it's not ready for review yet. Probably should have just dropped a message in discord. I have a demo at https://stampy-ui.ccstan99.workers.dev/ so we can test it on Firefox and mobile. I didn't see the blinking results on Firefox when I tested it on my machine but don't expect it to have magically fixed for people who were seeing that either.

ccstan99: OK, I set the debounce to 100ms for now.

ccstan99:
I also added userQuery along with the TF results as a start to addressing #67

app/routes/questions/add.tsx
```diff
@@ -5,7 +5,7 @@ import {redirect} from '@remix-run/cloudflare'
 import {addQuestion, loadAllQuestions, fetchJsonList, RelatedQuestions} from '~/server-utils/stampy'
 
 const getRelated = async (question: string): Promise<RelatedQuestions> => {
-  const url = `${NLP_SEARCH_ENDPOINT}/api/search?query=${question}`
+  const url = `${NLP_SEARCH_ENDPOINT}/api/search?query=${question}&showLive=0`
```
mruwnik, line 8: shouldn't these be for all questions, rather than just unpublished?

app/components/search.tsx
```diff
@@ -23,12 +23,11 @@ type SearchResult = Question & {
   url?: string
 }
 
-type WorkerMessage =
```
mruwnik, line 26: [NIT] it seems a bit iffy to have this as two separate types

ccstan99: Good point! I was so fixated on getting rid of the TypeScript errors that I didn't think through the fundamentals. Well, I tried this...
```
type WorkerMessage =
  | { status: string; numQs: number }
  | { searchResults: {title: string; pageid: string; score: number}[]; userQuery: string }
```
but it's still giving me these errors:
```
app/components/search.tsx:45:16 - error TS2339: Property 'status' does not exist on type 'WorkerMessage'.
  Property 'status' does not exist on type '{ searchResults: { title: string; pageid: string; score: number; }[]; userQuery: string; }'.
45       if (data.status === 'ready') {
                  ~~~~~~
app/components/search.tsx:49:16 - error TS2339: Property 'searchResults' does not exist on type 'WorkerMessage'.
  Property 'searchResults' does not exist on type '{ status: string; numQs: number; }'.
49       if (data.searchResults) {
                  ~~~~~~~~~~~~~
app/components/search.tsx:50:31 - error TS2339: Property 'searchResults' does not exist on type 'WorkerMessage'.
  Property 'searchResults' does not exist on type '{ status: string; numQs: number; }'.
50         setSearchResults(data.searchResults)
```
So I'm obviously missing something. To be honest, my understanding of TypeScript is purely by trial and error. How would you suggest to fix this properly?

app/routes/questions/add.tsx
ccstan99: Oops, I think I might have clobbered the work you did yesterday since I created this branch & made these changes earlier but didn't have time to continue working on this until now. I thought this function was for when the user is suggesting a new question. If that's the case, they should be requested only unpublished / non-live questions. Otherwise presumably, they should have picked a live / published answered from the previous screen. Anyway, I removed it.

mruwnik: won't it need `status=all`? i.e. I forgot about this url, so it's good that you found it :D

app/components/search.tsx
mruwnik: It think something like this should work:
```
type WorkerMessage = {
  searchResults?: {title: string; pageid: string; score: number}[]
  status: string
  numQs?: number
}
```

app/routes/questions/add.tsx
Aprillion: @mruwnik I also thought that this is for unpublished suggestions,, why do we want all, please?

mruwnik: this is what gets put into the possible duplicates column of https://coda.io/d/AI-Safety-Info_dfau7sl2hmG/Incoming-questions_su7CE#_luTjm
The unpublished questions for the dialog thingy that are to be bumped, are in `app/routes/questions/search.tsx`

mruwnik: There really should be some kind of central place/abstraction for all this :/

app/components/search.tsx
ccstan99: Ah, that did it, thanks! Aprillion tried explaining to me what the ? meant but I guess it didn't stick.

app/routes/questions/add.tsx
ccstan99: I think the user should only see and request unpublished questions in the dialog box. What do you think, @plex? I thought the return list was getting filtered out afterward and that's why there weren't always 5 listed.

mruwnik: that was the previous approach before you added filtering. Currently, there are 3 places that use search:
* the main search bar that uses tensor flow
* the dialog with unpublished questions that can be boosted -> this is in `app/routes/questions/search.tsx`
* an additional call to semantic search for all similar questions, when adding a new question -> this one here

ccstan99: It's way past my bedtime and I'm delirious... I'm not fully following the difference between the last 2. So, are the results from this list ever presented to the user? Or is it just one additional check before adding it to Coda? Anyway, I have an impending work deadline, so I'll have to leave this as is for now. If you need this fixed or merged sooner, feel free to finish it off. Otherwise unfortunately, I won't be able to get back to this again until later in the week. Thanks!

mruwnik: It's not even a check. The Incoming questions table in coda has a column for similar questions that are possible duplicates, to make it easier to decide how to handle new questions. This column is populated from semantic search, over the whole set of questions. Coda has a button that will go through the incoming questions and update the possible duplicates with values from semantic search. But the UI, when creating a new question, can also fill that field. So when a new question is created, the UI will additionally fetch a bunch of similar questions and set the possible duplicates directly.

This is a long-winded way of saying that no, the user will never see this list. It's all handled in the backend and only used when writing stuff to Coda

ccstan99: I knew you were doing some backend magic with Coda, so I'll trust that I should just stay out of your way. Thanks for clarifying so much for though! Otherwise, I totally missed that point. Perhaps you might consider checking if the score is at least above some threshold (i.e. >0.7?) to avoid cluttering the Coda column with rather dissimilar questions?

mruwnik (dismissed):

app/routes/questions/add.tsx
mruwnik: The threshold can be set as a parameter in the Coda formula. Currently it's 0.3, which is a bit low, but the idea was to have 5 for each item, and the reviewer can decide what to do with them.

app/components/search.tsx
mruwnik: yup, typescript is annoying in that way :D

Aprillion: typescript is nice, when all types are shared ... but the worker is native JavaScript file (executed as-is without compilation in the browser for now), so the TS type must match what the worker actually sends and there is no TS error if we forget to synchronize both

feel free to refactor to compile the workef rom TS you want to make it less annoying 😸

Aprillion: actually, the type doesn't match - following lines post message from the worker:
```js
23: self.postMessage({status: 'ready', numQs: questions.length})
39: self.postMessage({searchResults: []})
68: self.postMessage({searchResults, userQuery})
```

the correct type for that is:
```ts
type WorkerMessage = {
  status: 'ready'
  numQs: number
} | {
  searchResults: {title: string, pageid: string, score: number}[]
  userQuery?: string
}
```

and fixing the type will probably produce TS errors in this file => @ccstan99 let me know if you want me to fix all TS thingies myself

Aprillion (changes requested):

app/components/search.tsx
ccstan99: Yes, pretty please! I'd love to learn TS (and a bunch of other things) properly some day. I can definitely see its potential for beauty and power. Unfortunately, I don't have the bandwidth today.

ccstan99: @Aprillion Should I wait for you to take care of the TS thingies before merging? Or should we merge first and cleanup later? Otherwise, I'll need to find some time to go back to my trial and error approach with TS. Would you mind sending me that link that explains how those ? work in TS again? I forgot what they're even called. I remember it made sense initially, but unfortunately, old age & memory don't go well together.

Aprillion (approved):

Issue #88, opened 2022-08-14
Social preview for links to a single question


plexish:
e.g. ![image](https://user-images.githubusercontent.com/1039691/184558459-acbe4036-5ff0-4b6b-a45b-d8442633fb14.png)

This would be super nice for social media shares, especially once we have the copy direct link icon.

The title should be the title of the question, the body should be the first 100 characters of the answer. Ideally include " - Stampy" or " - AI Safety Info" for the respective sites in the title, but using " - Stampy's AI Safety Info" for both is ok if it's annoying to have it site-specific.

Aprillion:
ooh, server side rendering of social preview 🤔  CF workers don't have `canvas` AFAIK, so no OG:image, but the `title` and `description` should be possible if we add a root loader in https://github.com/StampyAI/stampy-ui/blob/42ad8c6ac623f625f2381d128ccd30425de9186b/app/root.tsx

PR #188, opened 2023-02-26
Add url previews

mruwnik:
image


app/root.tsx
```diff
@@ -3,20 +3,59 @@ import type {MetaFunction, LinksFunction, LoaderFunction} from '@remix-run/cloud
...
+import {loadQuestionDetail} from '~/server-utils/stampy'
 
-const title = 'Stampy'
-const description = 'AI Safety FAQ'
+/*
+ * Transform the given text into a meta header format.
+ *
+ * In practice, this means stripping out any HTML tags and limiting its length
+ */
+const metafyText = (text: string | null, defaultText: string) => {
```
Aprillion, line 14: eeh, can we find a better variable name? e.g. `makeSocialPreviewText`

app/root.tsx
```diff
@@ -3,20 +3,59 @@ import type {MetaFunction, LinksFunction, LoaderFunction} from '@remix-run/cloud
...
+  if (!text || text.length == 0) return defaultText
+
+  let cleaned = text
+  // Totally remove any P, A or IFRAME tags, but leave the contents. Any other
+  // tags will be ignored for now
+  const removals = [/<\/?p.*?>/gi, /<\/?a.*?>/gi, /<\/?iframe.*?>/gi]
```
Aprillion, line 20: I guess the regex is good for now, but when @plexish will test it more and asks for changes to the removals (or other modifications), it will be even better to add unit tests for the input/output examples to this function

IMHO this is the first big feature that the (_implicit manual regression testing during_) local development is unlikely to discover regression bugs

app/root.tsx
```diff
@@ -3,20 +3,59 @@ import type {MetaFunction, LinksFunction, LoaderFunction} from '@remix-run/cloud
...
+  cleaned = removals.reduce((str, regex) => str.replace(regex, ''), cleaned)
+
+  // This first random website I found claims that the description limit is 155,
+  // so keep it to that: https://www.contentkingapp.com/academy/meta-description/
+  if (cleaned.length > 155) cleaned = cleaned.slice(0, 152) + '...'
```
Aprillion, line 25: have you tested on Discord, Slack, Twitter, and Facebook yet? (what else do we care about?) we should aim for shortest supported by all (155 if fine if none of the sites shrinks it even more - I just don't want to force editors to test it on multiple sites for "unfortunate" shortening variations)

app/root.tsx
```diff
@@ -3,20 +3,59 @@ import type {MetaFunction, LinksFunction, LoaderFunction} from '@remix-run/cloud
...
+export const meta: MetaFunction = ({data}) => {
+  const title = metafyText(data?.question?.title, TITLE)
+  const description = metafyText(data?.question?.text, DESCRIPTION)
+  return {
+    title,
+    description,
+    'og:title': title,
+    'og:description': description,
+    'og:image': ogImage,
+    'twitter:image': ogImage,
```
Aprillion, line 55: a) looks like we have a bug in the og/twitter meta tags, the image doesn't work

b) don't use this image with generic preview, use the appropriate logo (e.g. passing props via `<Meta minLogo={minLogo} />` ... or if that doesn't work, we can use explicit <meta ...> without the MetaFunction)

Aprillion (changes requested):
FYI: Moving #88 to "In Progress" in https://github.com/orgs/StampyAI/projects/1/views/3 now that we have the Kanban board...

The volunteer team is growing, so let's try to keep PRs (or commits) well linked to issues, so we don't have to spend half of the meetings cleaning up task boards and to generally have better time in future git archeology when investigating bugs and trying to find out "why did we need this in the first place" 😅

app/root.tsx
mruwnik: Discord and Slack work, I don't have Twitter or Facebook, so can't check them

Aprillion: twitter doesn't work (probably never worked in previous version either, not a new bug, but worth fixing nevertheless):

image

error message on https://cards-dev.twitter.com/validator `ERROR: No card found (Card error)` does not sound very useful, but feel free to use the validator for any fixing ideas

linkedin doesn't like it either:
image

facebook looks half-way there:
image

Aprillion: even on Discord, the image is missing (which is probably why you didn't spot that it's a wrong image)

Aprillion: so, @plexish wants to triple the text in Discord (discussed in https://discord.com/channels/677546901339504640/758062805810282526/1079418708394512455) => feel free to increase the limit even if it would get shortened on other platforms, not a big deal for me to keep the text length consistent

tayler6000:
> FYI: Moving #88 to "In Progress" in https://github.com/orgs/StampyAI/projects/1/views/3 now that we have the Kanban board...
> 
> The volunteer team is growing, so let's try to keep PRs (or commits) well linked to issues, so we don't have to spend half of the meetings cleaning up task boards and to generally have better time in future git archeology when investigating bugs and trying to find out "why did we need this in the first place" 😅

Thank you for finding the issue for me.  I try to login every day to make sure all PRs have the appropriate tags, are linked to existing issues if they exist, and if not are in the project.  Thank you for taking off some of my workload 😄

Aprillion (approved):

Issue #187, opened 2023-02-26
Dialog box too skinny


ccstan99:
Dialog box after user clicks on "I'm asking something else" is too skinny
![dialog](https://user-images.githubusercontent.com/38997995/221405148-9d44ecb8-186d-4380-a068-d45a6d8d226c.jpg)

Shape & dimensions be something like this
![correct](https://user-images.githubusercontent.com/3942390/220209009-a39253bf-2892-419d-a882-7fc99cfcd080.png)

PR #189, opened 2023-02-26
Ensure new questions dialog has min width

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/187

### DIFF
--- a/app/components/search.tsx
+++ b/app/components/search.tsx
@@ -236,7 +236,6 @@ const ShowMoreSuggestions = ({
   } else if (extraQuestions.length === 0) {
     return (
       <Dialog onClose={onClose}>
-        <div style={{minWidth: '30em'}}></div>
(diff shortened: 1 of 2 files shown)


app/components/search.tsx
```diff
@@ -236,6 +236,7 @@ const ShowMoreSuggestions = ({
   } else if (extraQuestions.length === 0) {
     return (
       <Dialog onClose={onClose}>
+        <div style={{minWidth: '30em'}}></div>
```
mruwnik, line 239: this fixes the issue, but would it be better to have it in the css file?

Aprillion: if `<Dialog className=...` works, can you add a new class to CSS?

Aprillion (changes requested):

Aprillion (approved):

PR #193, opened 2023-03-04
Url to iframe tests

jrhender:
This PR:
- Adds `jest` dependencies to use `jest` for tests. Jest selected because it provides full functionality for units tests (e.g. test runner, mocking, assertions) in a single package
- Adds npm script to execute tests
- Adds `npm run test` to `pr-check.yml`
- Adds note to README on running tests
- Adds simple tests for `urlToIframe` function


.github/workflows/unit-test.yml
```diff
@@ -0,0 +1,15 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: unit-test
+        run: |
+          npm ci
+          npm run test
```
Aprillion, line 15: please move the `npm run test` into `pr-check.yml` file, not a new GH action

.gitignore
```diff
@@ -9,3 +9,6 @@ node_modules
 
 .DS_store
 wrangler.toml
+
+## vscode
+launch.json
```
Aprillion, line 14: unless you store any secrets in this file, I would be fine including it in git for new developers to have better VSCode integration

app/server-utils/url-to-iframe.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+    const input = `<a href="https://aisafety.world/">this map</a>`
+    expect(urlToIframe(input)).toBe(input);
+  });
+
+  it('should convert an anchor tag whose href matches tag content', () => {
+    const input = `<a href="https://aisafety.world/">https://aisafety.world/</a>`
+    const output = `<iframe src="https://aisafety.world/" sandbox="allow-scripts allow-same-origin"/>`
+    expect(urlToIframe(input)).toBe(output);
+  });
+})
```
Aprillion, line 14: did you run prettier for the new files (`npm run lint`)? I believe it should have added newline at the end of files automatically...

otherwise I like the first test 👍

Aprillion (approved):

.github/workflows/unit-test.yml
jrhender: Yes, you're right, I think that is simpler for now.
Done here: https://github.com/StampyAI/stampy-ui/pull/193/commits/fed9bc3f2749d6f28bdf354ff0560fe003193ddd

.gitignore
jrhender: Ok, sounds good. Added here: https://github.com/StampyAI/stampy-ui/pull/193/commits/0307b767e45c247aed533982e6d45fa958df9661

app/server-utils/url-to-iframe.spec.ts
jrhender: Thanks, yes, you were right, it needed a prettier run https://github.com/StampyAI/stampy-ui/pull/193/commits/bbb7a85290548f2b6f459252704e1b55ae1efa26

README.md
```diff
@@ -65,6 +65,12 @@ Once that's done, you should be able to test and deploy your app!
 $ npm run dev
 ```
 
+## Test
+
+```sh
+$ npm run test
+```
```
Aprillion, line 72: 👍

Aprillion (approved):

.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
```
mruwnik, line 1: should this be here? It looks like an editor specific file. Could you also add it to .gitignore?

app/server-utils/url-to-iframe.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+describe('urlToIframe', () => {
+  describe('should not convert a markdown link', () => {
+    it('whose text does not match url', () => {
+      const input = `[this map](https://aisafety.world/)`
+      expect(urlToIframe(input)).toBe(input)
+    })
+
+    it('whose text matches url but is not whitelisted', () => {
+      const url = 'https://example.com/'
+      const input = `[${url}](${url})</a>`
```
mruwnik, line 12: the trailing `</a>` should probably be removed

app/server-utils/url-to-iframe.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+      expect(urlToIframe(input)).toBe(input)
+    })
+  })
+
+  describe('should convert a markdown link', () => {
+    it('whose text matches url and is whitelisted', () => {
```
mruwnik, line 18: could you also test these?:
* `[https://aisafety.world/?bla=bla](https://aisafety.world/?bla=bla)`
* `[https://aisafety.world/](https://aisafety.world/?bla=bla)`
* `[https://aisafety.world/?bla=bla](https://aisafety.world/)`
* `[http://aisafety.world/](http://aisafety.world/)`
* `[aisafety.world](http://aisafety.world/)`
* `[aisafety.world/](aisafety.world/)`

README.md
mruwnik: 👍

mruwnik (changes requested):

mruwnik:
This is really useful - good work!

.vscode/launch.json
mruwnik: ah. I see @Aprillion wants this - disregard this comment :D

app/server-utils/url-to-iframe.spec.ts
jrhender: Yes, certainly. Thanks for thinking of all these cases.

What should the expected results be for each of them in your view? Should all of them convert into the iframe or just some of them?

I went back and looked at the GH issue for this iframe (https://github.com/StampyAI/stampy-ui/issues/140) and @plexish was saying that "If a full plain URL from the whitelisted domains is in the markdown, it should be transformed into a large (~screen height, full answer width) iframe of that URL."
Re-reading that, I wonder if my initial understanding and implementation isn't quite correct. The current implementation is considering the `href`/`()` portion of the link but maybe this is not relevant. Maybe it just matters that the text/`[]` portion is a whitelisted URL.

With this new understanding, I think the expected behaviour should be:
1. `[https://aisafety.world/?bla=bla](https://aisafety.world/?bla=bla)` -> should render iframe with href of `https://aisafety.world/?bla=bla`
1. `[https://aisafety.world/](https://aisafety.world/?bla=bla)` -> should render iframe with href of `https://aisafety.world/`
1. `[https://aisafety.world/?bla=bla](https://aisafety.world/)`  render iframe with href of `https://aisafety.world/?bla=bla`
1. `[http://aisafety.world/](http://aisafety.world/)` -> should render iframe with href of `https://aisafety.world/`
1. `[aisafety.world](http://aisafety.world/)` -> should **not** render iframe because it's missing URI scheme
1. `[aisafety.world/](aisafety.world/)`  -> should **not** render iframe because it's missing URI scheme

Does the above seem right?

mruwnik: I'd personally go with both having to be identical. But what you propose is also totally reasonable. The main thing is to have a consistent view

plexish: I'm okay with either of these schemes a long as it's documented somewhere the editors can read like the formatting question.

jrhender: Ok, I think "both having to be indentical" is the easier route for now and I ideally don't want to change the implementation in an "adding tests" PR.

That said, the first test case you suggested doesn't create an iframe with the current implementation. I think the query string breaks it.
> `[https://aisafety.world/?bla=bla](https://aisafety.world/?bla=bla)` -> should render iframe with href of `https://aisafety.world/?bla=bla`

I made it a failing test case: https://github.com/StampyAI/stampy-ui/pull/193/commits/97390534ac31e9f464066b639997ae36285b96e0
We can come back and add support for query strings in a separate PR, IMO

mruwnik: nice! @Aprillion would you prefer this to be fixed now, or do you think that urls with query strings should work as they do now? 
@jrhender if you want to fix it, the regex should be `const matches = text.matchAll(/\[(https?:\/\/[\.\w]+?\/?[^\]]*?)\]\(\1\)/gi)`, as it is in the comment

Aprillion: I would prefer to fix it, that's the purpose of tests, no strong opinion between following options:
* a) this PR with [it.failing(...)](https://jestjs.io/docs/api#testfailingname-fn-timeout), and the PR author can file a GH issue with the bug and then someone can fix code in next PR, or
* b) 2 separate commits in this PR so that tests are passing before we merge, or
* c) 1 commit with a passing test and fixed code at the same time

just no merging of red tests, please

up to @jrhender what option you feel most comfortable with, though TBH, option b) sounds most straight forward (if the code from above comment fixes the problem without making the other tests unhappy)

jrhender: Ah I didn't now about `it.failing`, that's nifty.

Ok, I moved querystring case to be expected to pass and adjusted regex as per mruwnik's tips https://github.com/StampyAI/stampy-ui/pull/193/commits/566a78fc4f18e2ffde2246e583222f8022f7298c

jrhender: > as it's documented somewhere the editors can read like the formatting question.

@plexish I'm glad to document this. Do you know where a good place to do this might be? Is there a Q&A writing guide perhaps?

mruwnik (dismissed):

jrhender:
Ah, I see I need to run prettier

app/server-utils/url-to-iframe.spec.ts
plexish: yep write it in a comment there and I'll include it @jrhender

PR #197, opened 2023-03-11
Fun logo on stampy.ai,  inifinite scroll less trigger happy

### LINKED ISSUE #157: Switch stampy.ai main domain to point to the UI

Aprillion:
* npm audit fix
* use fun logo on stampy.ai top domain, fixes #157
* make infinite scroll less trigger happy
   * load only if user actually scrolls (for direct links to a question that have small content)
   * preserve url parameters in cache reload
   * stop reloading all pages when the last one is reached

### DIFF
--- a/app/components/infiniteScroll.tsx
+++ b/app/components/infiniteScroll.tsx
@@ -1,58 +1,43 @@
 import {ReactNode, useState, useRef, useCallback} from 'react'
-import {Question} from '~/server-utils/stampy'
 
 type Props = {
-  children: ReactNode[]
-  fetchMore: () => Promise<Question[]>
-  loadingRenderer?: (loading: boolean) => ReactNode
+  children: ReactNode
+  fetchMore: () => Promise<string | null>
 } & JSX.IntrinsicElements['div']
 
-export default function InfiniteScroll({
-  children,
-  fetchMore,
-  loadingRenderer = (loading: boolean) => loading && <div className="loader"></div>,
-  ...props
-}: Props) {
-  const [pageNum, setPageNum] = useState(0)
-  const [hasMore, setHasMore] = useState(true)
+export default function InfiniteScroll({children, fetchMore}: Props) {
   const [loading, setLoading] = useState(false)
 
-  const getMore = useCallback(async () => {
-    if (loading || !hasMore) return
+  // must use stable callback without depencencies, otherwise it will create many instances of IntersectionObserver
+  const fetchMoreRef = useRef(fetchMore)
+  fetchMoreRef.current = fetchMore
+  const lastElementRef = useCallback((node: HTMLDivElement) => {
+    if (!node) return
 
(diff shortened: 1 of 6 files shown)


app/root.css
```diff
@@ -62,7 +63,7 @@
 }
 
 html {
-  height: 100%;
+  height: 101%; /* make sure infinite scroll trigger is below initial screen */
```
mruwnik, line 66: nice!

app/root.tsx
```diff
@@ -67,10 +67,10 @@ export const meta: MetaFunction = ({data}) => {
...
 
 export const loader = async ({request}: Parameters<LoaderFunction>[0]) => {
-  const isDomainWithLogo = request.url.match(/ui.stampy.ai/)
-  const isLogoForcedOff = request.url.match(/minLogo/)
-  const isLogoForcedOn = request.url.match(/funLogo/)
-  const minLogo = isDomainWithLogo ? !!isLogoForcedOff : !isLogoForcedOn
+  const isDomainWithFunLogo = request.url.match(/stampy.ai|localhost/) // min logo by default on aisafety.info and 127.0.0.1
+  const isFunLogoForcedOff = request.url.match(/minLogo/)
+  const isFunLogoForcedOn = request.url.match(/funLogo/)
+  const minLogo = isDomainWithFunLogo ? !!isFunLogoForcedOff : !isFunLogoForcedOn
```
mruwnik, line 73: line 47 will probably also need to be changed :/

mruwnik (approved):

app/components/infiniteScroll.tsx
```diff
@@ -1,58 +1,43 @@
...
+
+        const nextPageLink = await fetchMoreRef.current()
+        if (!nextPageLink) {
+          observer.disconnect()
+        }
+        setLoading(false)
+        loadingInternal = false
       }
-    },
-    [getMore]
```
Aprillion, line 49: @mruwnik since the dependencies changed almost in every rerender, it didn't make much sense to wrap it in `useCallback`, it was called from the callback ref quite a lot, disconnecting and reconnecting new observers .. not a big deal, but logic for `nextPage` was wrong (the `fetchMore()` actually returned questions array, and it wasn't empty on the last page, so it just started to reload everything from the first page again even when all 197 questions were already displayed...)

so together with fixing that logic, I also simplified it to single stable function that uses mutable refs and only creates 1 observer on mount, then disconnects the observer when there is no next page

app/root.tsx
Aprillion: huh? not sure what you mean, it's currently `${url.origin}/favicon-512.png` so not depending on these conditions...

Aprillion: if there is a feature request to make the social preview logo dynamic too, we can do that in future PR, not blocking merge ;)

PR #198, opened 2023-03-12
fix infinite scroll on Chrome

Aprillion:


Aprillion:
it's a CSS-only change, let me merge right away to fix the issue in Chrome + feel free to comment if you have any concerns that we should improve something here

PR #201, opened 2023-03-19
Fix double scroll on mobile and related questions filter

Aprillion:

### DIFF
--- a/app/hooks/useQuestionStateInUrl.ts
+++ b/app/hooks/useQuestionStateInUrl.ts
@@ -1,8 +1,14 @@
 import {useState, useRef, useEffect, useMemo, useCallback} from 'react'
 import type {MouseEvent} from 'react'
 import {useSearchParams, useTransition} from '@remix-run/react'
-import {Question, QuestionState, RelatedQuestions, PageId} from '~/server-utils/stampy'
-import {fetchOnSiteAnswers} from '~/routes/questions/allOnSite'
+import {
+  Question,
+  QuestionState,
+  RelatedQuestions,
+  PageId,
+  QuestionStatus,
+} from '~/server-utils/stampy'
+import {fetchAllQuestionsOnSite} from '~/routes/questions/allQuestionsOnSite'
 import {
   processStateEntries,
   getStateEntries,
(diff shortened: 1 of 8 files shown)


app/hooks/useQuestionStateInUrl.ts
```diff
@@ -116,6 +123,7 @@ export default function useQuestionStateInUrl(minLogo: boolean, initialQuestions
 
     const onSiteAnswers = onSiteAnswersRef.current
     const onSiteSet = new Set(onSiteAnswers.map(({pageid}) => pageid))
+    console.debug(relatedQuestions, onSiteSet)
```
mruwnik, line 126: should this be here?

app/server-utils/stampy.ts
```diff
@@ -283,13 +283,15 @@ export const loadTag = withCache('tag', async (tagName: string): Promise<Tag> =>
 
   const questions = await loadAllQuestions()
   const nameToId = Object.fromEntries(
-    questions.data.filter((q) => q.status == QuestionStatus.LIVE).map((q) => [q.title, q.pageid])
+    questions.data
+      .filter((q) => q.status == QuestionStatus.LIVE_ON_SITE)
```
mruwnik, line 287: I wonder if there should be a general filter function to handle this? I'm guessing there will be more such situations in the future :/

app/root.css
```diff
@@ -181,7 +181,6 @@ button.icon-link,
 a:visited.icon-link {
   font-size: 0.6rem;
   text-transform: uppercase;
-  white-space: nowrap;
```
mruwnik, line 184: this is all black magic for me, but I assume it works

mruwnik (approved):

app/root.css
Aprillion: it's half black magic to me too  🪄 

the use of `overflow-x: hidden` was triggering a change of default rule for `overflow-y: auto` causing `body` and `html` to be both scrollable on mobile screens 🙀 

...so I had to find a different alternative for overflow-x and I pinpointed this issue as causing the horizontal scrolling issue (a few px of overflow-x on <1000px window) by disabling 1 rule at a time for "a few minutes"

...and then I had to figure out another nowrap equivalent, so I changed the previous `flex: 0` (with implied flex-basis as "0%") to `flex: none` (with implied flex-basis as "auto" which defaults to the width of the child before wrapping - at least when some particular constellation of stars is most to our code)

🤞 hopefully no further chain reaction 🍀

app/hooks/useQuestionStateInUrl.ts
Aprillion: not really, but I no longer care today 🤷‍♀️

feel free to remove in next PR (or I will remove myself when I notice next time)

app/server-utils/stampy.ts
Aprillion: sure, but it sounded like "a future problem" to me right now 😅

PR #216, opened 2023-04-22
Show only onSite related questions

### LINKED ISSUE #214: WIP questions showing as related
At this URL
https://aisafety.info/?state=1001_
there are WIP follow up questions show up underneath.

Aprillion:
Fix #214

### DIFF
--- a/app/hooks/useQuestionStateInUrl.ts
+++ b/app/hooks/useQuestionStateInUrl.ts
@@ -200,7 +200,7 @@ export default function useQuestionStateInUrl(minLogo: boolean, initialQuestions
       const onSiteAnswers = onSiteAnswersRef.current
       if (onSiteAnswers.length === 0 && relatedQuestions.length > 0) {
         // if onSiteAnswers (needed for relatedQuestions) are not loaded yet, wait a moment to re-run
-        setTimeout(() => toggleQuestion(questionProps, options), 500)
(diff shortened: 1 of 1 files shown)


app/hooks/useQuestionStateInUrl.ts
```diff
@@ -200,7 +200,7 @@ export default function useQuestionStateInUrl(minLogo: boolean, initialQuestions
       const onSiteAnswers = onSiteAnswersRef.current
       if (onSiteAnswers.length === 0 && relatedQuestions.length > 0) {
         // if onSiteAnswers (needed for relatedQuestions) are not loaded yet, wait a moment to re-run
-        setTimeout(() => toggleQuestion(questionProps, options), 500)
+        setTimeout(() => toggleQuestion(questionProps, options), 200)
```
mruwnik, line 203: should this have some kind of back-off mechanism? Or at least only try n times?

mruwnik (approved):

app/hooks/useQuestionStateInUrl.ts
Aprillion: good question, but it's not doing any requests, so even if there is a problem and `onSiteAnswers` are never loaded, doing an `if` comparison every 200ms won't fry any eggs => not a problem IMHO

Issue #217, opened 2023-04-22
Error from NLP_SEARCH_ENDPOINT


Aprillion:
I am getting `JSON.parse: unexpected character at line 1 column 1 of the JSON data` error from Remix when I click `I'm asking something else` that fetches the `${NLP_SEARCH_ENDPOINT}/api/search?query=${question}&top=5&showLive=` thingy while I was fixing previous issue...

Based on my env variables, I tried to check https://stampy-nlp-t6p37v2uia-uw.a.run.app/api/search?query=corr&top=5&showLive=0 but that seems to work fine and there is no better debugging set up yet from Remix error handling as far as I could see => needs more investigation that I have time right now.

PR #219, opened 2023-04-22
Catch errors from search endpoint

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/217

image

### DIFF
--- a/app/routes/questions/search.tsx
+++ b/app/routes/questions/search.tsx
@@ -15,6 +15,10 @@ export function search(question: string, onlyLive: boolean) {
     onlyLive ? 1 : 0
   }`
   return fetch(url).then(async (response) => {
+    if (response.status != 200) {
+      console.error(await response.text())
+      return new Response('Error while searching for questions', {status: 502})
+    }
     return await response.json()
   })
 }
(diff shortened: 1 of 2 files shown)


mruwnik:
should the error be nicely formatted?

app/routes/questions/search.tsx
```diff
@@ -15,6 +15,10 @@ export function search(question: string, onlyLive: boolean) {
     onlyLive ? 1 : 0
   }`
   return fetch(url).then(async (response) => {
+    if (response.status != 200) {
+      console.error(await response.text())
+      return new Response('Error while searching for questions', {status: 502})
```
mruwnik, line 20: I considered returning what the downstream server returns, but that's usually random HTML, so decided against it

Aprillion: please do - we don't have to show it in UI, but at least make it visible in network tab of the browser

`console.error` from here is going to CloudFlare, not very useful for everyday debugging 😅

Aprillion: (and no need for nice formatting of error in UI)

Aprillion (approved):

PR #235, opened 2023-05-14
Stop magic replacing gdoc links if known

mruwnik:


Aprillion (commented):
Looks fine from code review, have you tested a few answers that worked before also work after the GDocs parsing change? (in some non-production CF worker or at least localhost)

mruwnik:
yup. I'll also wait a bit to see if it's still working before I merge this - it was mainly a matter of just getting it out

Aprillion (approved):

PR #237, opened 2023-05-17
Use alternative phrasings in search

mruwnik:
https://github.com/StampyAI/stampy-nlp/issues/13


ccstan99 (commented):
Could be potentially confusing, at least awkward, for users to select an alternate phrasing but the actual (original) question displayed is different, i.e. selecting "Why can't we simply stop developing Al?" displays "Why don't we just not build AGI if it's so dangerous?" I think we might want to demo this at next dev meeting to discuss.

Screenshot 2023-05-19 at 1 44 20 AM

Aprillion (approved):

Issue #174, opened 2023-02-19
Log all searches


mruwnik:
It would be useful to have a collection of what gets searched.
* saved to coda?
* cloudflare + export?

ccstan99:
All calls sent through the semantic search /api/search are logged as part of the Google Cloud Run. That can be exported periodically. Currently, the in-browser TF searches are not tracked. So we can either log that or wait until it gets integrated with the /api/search.

mruwnik:
When is it likely to get integrated? That does sound like a much better solution, on multiple levels

ccstan99:
I hope to have a better idea after testing out the responsiveness next weekend. If we can move it over, shouldn't take too long.

tayler6000:
Idea approved in 26 Feb 2023 Office Hours

diziet:
If we implement a feature like this, I would suggest logging these queries in such a way that they can not be tied to any PII (personally identifiable information) from the user and additionally create a Terms of Service and Privacy page that unambiguously and clearly to the user spells out how the data is used. The ideal storage functions would make us compliant with the variety of privacy regulation around the world. See https://github.com/StampyAI/stampy-ui/issues/81 for related discussion

Aprillion:
@diziet if you've done some research on best TOS that we could reuse, can you please link in here? If we choose to go that way on the next dev call, it might be useful to have all info ready for implementation, thanks! ;)

mruwnik:
can be viewed at https://console.cloud.google.com/logs/query;query=resource.type%20%3D%20%22cloud_run_revision%22%0Aresource.labels.service_name%20%3D%20%22stampy-nlp%22%0Aresource.labels.location%20%3D%20%22us-west1%22%0A%20severity%3E%3DDEFAULT%20queries;cursorTimestamp=2023-05-24T15:12:02.477604Z?project=stampy-nlp

PR #241, opened 2023-05-22
Log searches to the NLP endpoint

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/174

This is quite complicated, as I wanted to only log actual queries, rather than each button press. One way around this would be to only log stuff once the user presses enter, but I'm guessing that would massively limit the number of searches logged. So I settled on this, which amounts to:

* if the current search value is significantly different from the previous one (i.e. neither is a substring of the other), then log the search value
* if the search bar hasn't been changed for the last 4 seconds, log the search term, as the user has probably either found what they were looking for, or have given up

### DIFF
--- a/app/components/search.tsx
+++ b/app/components/search.tsx
@@ -75,6 +75,7 @@ export default function Search({onSiteAnswersRef, openQuestionTitles, onSelect}:
     const value = rawValue.trim()
     if (value === searchInputRef.current) return
 
+    logSearch(value)
     setLoading(true)
     searchInputRef.current = value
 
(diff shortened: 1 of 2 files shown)


app/components/search.tsx
```diff
@@ -378,3 +379,41 @@ const normalize = (question: string) =>
  * Sort function for the highest score on top
  */
 const byScore = (a: SearchResult, b: SearchResult) => b.score - a.score
+
+/**
+ * Handle flushing searched phrases to the NLP logger endpoint
+ */
+let prevSearch = ''
+let lastTimestamp = 0
```
mruwnik, line 387: I very much dislike these variables here, but I couldn't come up with a better way of keeping state. It's fine from a concurrency point of view, as the `logSearch` function should be called sequentially, anyway

app/components/search.tsx
```diff
@@ -378,3 +379,41 @@ const normalize = (question: string) =>
...
+
+const logSearch = (value: string) => {
+  setTimeout(shouldFlushSearch(value, prevSearch), 4000)
```
mruwnik, line 390: this is a potential heisenbug, as it's possible for the flush calls to come in the wrong order? Should be fine?

Aprillion (changes requested):

app/components/search.tsx
mruwnik: The search debounce is 100ms, which is not much. I'd prefer not to have logs full of things like the following:
```
Wh
Why
Why d
Why do
Why do I
Why do I ty
Why do I typ
```
This is good when searching, as the results should update often, but in the case of logging searches, it seems overkill - better to just log things which are relatively complete (whatever that means :D).

Debouncing will help with avoiding the multiple entries, at the cost of filtering out people who rapidly write in several queries before finding what they're looking for. I don't know what would be best

ccstan99: > better to just log things which are relatively complete

Is it possible to only log searches when the user clicks outside of search bar, i.e. selects something from dropdown menu or just abandons search? It'd be interesting to log the query as well as selection or abandonment if possible.

app/routes/questions/log.tsx
```diff
@@ -0,0 +1,8 @@
+import type {ActionArgs} from '@remix-run/cloudflare'
+
+export const action = async ({request}: ActionArgs) => {
+  const {query, name} = await request.json()
+  return await fetch(
+    `https://stampy-nlp-t6p37v2uia-uw.a.run.app/api/log_query?name=${name}&type=UI&query=${query}`
```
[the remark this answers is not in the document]
ccstan99: Logging the domain seems like a good idea!

app/components/search.tsx
Aprillion: hmm, 100ms is quite short ... when/why did we change it from 300 / 500ms like we had before?

in any case, I still vote that whatever is good for the search feature should also be good for logging - and if not, we need tests for the logging correctness

ccstan99: Updated late Feb because people were complaining about lag:
Why - https://github.com/StampyAI/stampy-ui/issues/175
When - https://github.com/StampyAI/stampy-ui/pull/186/files/7ffdf8becbf424eb155ef198d0c518b5ac4ea354

Aprillion (approved):

Issue #238, opened 2023-05-18
Feature Request: Ability to rehide text after [See more...] is clicked


markovial:
Currently in answers the see more just disappears after clicking. It would be nice if this was more like a toggle so that the reader can go back to what is considered the "core information" which appears before the tag.

Basically, if I click see more... and what shows up is too much information that I don't care about, I should be able to delineate between what the original author of the answer considered to be core info, and what was just extra see more info that I can hide again, if that makes sense...

[Here](https://aisafety.info?state=8HIA_) is an example of a random answer we a see more tag for easy reference of how it currently works.

plexish:
Is this needed given that you can collapse the whole answer anyway? The old wiki-based version had this, but it never felt super useful to me, and the hide link felt kinda cluttery. Open to more opinions or reasoning though.

tayler6000:
I think it's pretty standard UI that where there is a `see more`, there is a `see less`.  I'm also willing to get the opinions of others though.

plexish:
@ccstan99 do you have a quick take on this?

ccstan99:
Yes, if there's `see more`, there should be `see less`. It's both or none. Having just one without the other is indeed odd. 

Ideally, from a big picture point of view, it'd be better to cull out long answers so we shouldn't we need this at all. If it isn't "super useful", let's remove it entirely.

markovial:
I mean there are a bunch of things that I would not consider "super useful" when trying to just get the intuition of a topic across, but are still "useful" enough to distill.

The entire point of the see more tag in my mind was that if the reader has become interested in the thing enough, there is more that they could read about it which does not involve them having to go to an overly technical research paper or blog post.

plexish:
Okay, we can have both then.

My guess is it's pretty useful to have the option for extra detail in some case though?

PR #243, opened 2023-05-24
Show "See less" after toggling more

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/238

### DIFF
--- a/app/root.css
+++ b/app/root.css
@@ -643,9 +643,14 @@ a:focus-visible,
   border-radius: 2px;
 }
 
-a.see-more:not(.visible) + div.see-more-contents,
-a.see-more.visible {
+a.see-more:not(.visible) + div.see-more-contents {
   display: none;
+}
+a.see-more:after {
+  content: 'See more...';
+}
+a.see-more.visible:after {
+  content: 'See less';
 }
 
 .error {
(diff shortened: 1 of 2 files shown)


app/root.css
```diff
@@ -643,10 +643,15 @@ a:focus-visible,
...
-a.see-more:not(.visible) + div.see-more-contents,
-a.see-more.visible {
+a.see-more:not(.visible) + div.see-more-contents {
   display: none;
 }
+a.see-more:after {
+    content: "See more...";
+}
+a.see-more.visible:after {
+    content: "See less";
```
mruwnik, line 653: would it be better to do this in JS rather than with CSS?

app/server-utils/stampy.ts
```diff
@@ -199,7 +199,7 @@ const renderText = (text: string | null): string | null => {
   const wrapInDetails = ([chunk, ...rest]: string[]): string => {
     if (!rest || rest.length == 0) return chunk
     return `<div>${chunk}<div>
-           <a href="" class="see-more">See more...</a>
+           <a href="" class="see-more"></a>
```
Aprillion, line 202: hm, didn't pretter autoformat this into `<a ... />`? I guess no one is going to need to copy the text, so I don't mind doing it in CSS 🤷  .. maybe even better to make the long answer copy-able withtout `See less` 🦩

Aprillion (approved):

app/server-utils/stampy.ts
mruwnik: hmm. Turns out that if you copy the text it skips the `See less`, as that's added by CSS. So this works automagically :D

Issue #25, opened 2022-06-30
Make related questions which are already on the page relocate to collapsed in the sliding in place


plexish:

mruwnik:
This should probably also handle the case where opening a related question doesn't reshuffle them, leaving the list looking like any later related questions are part of the opened question, i.e.:
<img width="1070" alt="image" src="https://user-images.githubusercontent.com/3942390/225761077-13901f38-46a7-42fd-94a8-53dbd2627358.png">

Changes to:
<img width="1040" alt="image" src="https://user-images.githubusercontent.com/3942390/225761164-17675f18-2f48-4d74-b91a-87195971a191.png">

LeMurphant:
No longer relevant with new UI

PR #245, opened 2023-05-25
Relocate related questions

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/25

* when expanding a related question, move it below all other related questions (otherwise it just appears as a top level question with a random set of the parent question's related questions as its own related questions (too much recursion in that description...))
* pull in expanded related questions when opening a questions with related ones

### DIFF
--- a/app/hooks/stateModifiers.tsx
+++ b/app/hooks/stateModifiers.tsx
@@ -23,6 +23,11 @@ export const getStateEntries = (
     )
   ).filter(makeUniqueChecker())
 
+export const entryState = (state: StateString, pageid: PageId): QuestionState =>
+  getStateEntries(state)
+    .filter(([entryPageid]) => entryPageid === pageid.toString())
+    .map(([pageid, state]) => state)[0]
+
 export const removeRelated = (questions: StateEntry[]): StateEntry[] =>
   questions.filter((i) => i[1] != QuestionState.RELATED)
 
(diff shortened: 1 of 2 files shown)


app/hooks/stateModifiers.tsx
Aprillion, line 1: I'm afraid this still needs some UX improvements for questions with multiple related links like http://127.0.0.1:8787/?state=6178- .. opening one of the related links moves it below screen and also moves the parent question below screen, very confusing IMHO (especially when the parent is open and everything just jumps) .. not sure how to improve to correctly handle all important edge cases 😅

Aprillion (changes requested):

app/hooks/stateModifiers.tsx
mruwnik: Totally agree with that. @ccstan99 @plexish any ideas how it should work?

ccstan99: My initial thought is to have the related questions as a toggle list. 
- Only one list of related questions is open at any given time
- When a related question is selected, all the other related questions are collapsed into the toggle list.
There will be some unexpected movement and more bookkeeping of which questions are on screen & expanded, but we should be sure to scroll so the answer that was just recently expanded is in view. Unfortunately, I don't have access to image editing tools right now but can try to draw a diagram later if needed. 

I'm not loving this approach though. We might want to discuss further at the next dev call to see if there are other ideas and opinions before implementing.

Slightly separate issue based on what Aprillion pointed out: Do we really want an answer with 22 related questions now that we support tags? We might need a discussion on the philosophy behind related questions and how exhaustive vs suggestive they're intended to be.

Aprillion: I like the idea to scroll to the just-expanded answer and agree there should be fewer related questions per 1 parent.

If we move the next related question above the current parent, we might get away with the bookkeeping 🤔 it will look like the previous related questions are also related to the newly expanded parent, but that might be good enough for our "reading list" feature for now and we can complicate it later if needed 😅 

https://excalidraw.com/#room=b1392834d262e6d9a063,37R-5A0STeSAEaxqwmAg0A
image

PR #248, opened 2023-05-26
Handle glossary items

mruwnik:
image

image


app/routes/index.tsx
```diff
@@ -21,7 +21,8 @@ import {reloadInBackgroundIfNeeded} from '~/server-utils/kv-cache'
 export const loader = async ({request}: Parameters<LoaderFunction>[0]) => {
   try {
     const initialQuestionsData = await loadInitialQuestions(request)
-    return {initialQuestionsData}
+    const glossaryData = await loadGlossary(request)
```
Aprillion, line 24: Hm, I don't like fetching all answers from Coda before sending the first byte to the client 🤔 

Can you reuse `onSiteAnswersRef` in a way similar to how `<Search ...` component used the same lazy-loaded data?

mruwnik: I initially tried that way. The problem is with the contents, which don't get loaded until a given question is fetched. So this would require either doing some magic to load questions when someone hovers over a definition, or prefetching the questions that are used for the glossary. Coda's filtering options are rudimentary, to say the best, so the best alternative I could think of would be to first fetch the glossary tag, then get all questions that it uses, which would quickly blow up :/
My assumption here is that the glossary will be cached after the first invocation, so should be fast after that

app/routes/questions/$question.tsx
```diff
@@ -79,6 +81,7 @@ export function Question({
   useEffect(() => {
     if (text == null && !isLoading.current) {
       isLoading.current = true
+      // This is where the actual contents of the question get fetched from the backend
```
mruwnik, line 84: I added this comment here, as every couple of weeks I spend 30 minutes trying to find where the questions get loaded :D

Aprillion: We should load content of initial questions already with page load, so this is not the only place to load actual contents...

And I think someone changed Coda tables and/or someone is using wrong table id that Coda admins didn't want us to use, and we load all answer text while loading all 700 question titles and then that request is not cache-invalidated somehow, so "it's more complicated".

Aprillion: And direct links with question id were supposed to load the contents right away + server-render it too, not to fetch it lazily from useEffect ... that might also need some investigation if it won't be "auto-resolved" by fixing the cache issues 😓

Aprillion (approved):

app/routes/index.tsx
Aprillion: I still don't like making multiple requests from CF worker "backend" to Coda that fetch all data from all answers table, 1 should be enough and we should share the data for all features that need all data ... but I guess I can investigate duplicate request after (or ehm, by..) fixing the cache issues.

PR #260, opened 2023-06-07
Move semantic search to typescript

mruwnik:
adds a new `useSearch` hook to contain all the actual searching magic. The Search component now is just display stuff, and tfWorker.js can be removed, along with its communication magic. This currently doesn't work, but that's a different issue, that is fixed by [this PR](https://github.com/StampyAI/stampy-nlp/pull/28/files) (semantic search in general won't work until that gets merged in)


app/hooks/search.tsx
```diff
@@ -0,0 +1,171 @@
...
+/**
+ * Configure the search engine.
+ *
+ * This will download the embeddings for semantic search, but until that is set up, will
+ * use baseline search over the list of questions already loaded on the site.
+ * Searches containing only one or two words will also use the baseline search
+ */
+export const useSearch = (onSiteQuestions: MutableRefObject<Question[]>, numResults = 5) => {
+  const baselineSearch = makeBaselineSearcher(onSiteQuestions, numResults)
+  const [searchFn, setSearchFn] = useState(() => baselineSearch)
```
Aprillion, line 138: why calling `makeBaselineSearcher()` every render and then only use it once?

app/hooks/search.tsx
```diff
@@ -0,0 +1,171 @@
...
+        pageids.map((pageid: string, i: number) => ({pageid, title: questions[i]})),
+        tf.tensor2d(encodings),
+        numResults
+      )
+      // warm up model for faster response later
+      await searcher('What is AGI Safety?')
+      setSearchFn(() => searcher)
+    }
+    loadModel()
+  }, [numResults, onSiteQuestions])
```
Aprillion, line 157: what's `onSiteQuestions` doing here?

app/hooks/search.tsx
```diff
@@ -0,0 +1,171 @@
...
+    .replace(/[^\w ]|\b(?:an?|the?)\b/g, '')
+    .replace(/(\w{2})s\b/g, '$1') // cannot use lookbehind (?<=...) because not supported on Safari
+    .replace(/\s+|_|&\s*/g, ' ')
+    .trim()
+
+/**
+ * Construct a semantic search function, that will search over the provided questions
+ */
+const semanticSearcher = async (questions: Question[], encodings: Tensor2D, numResults: number) => {
+  const model = await load()
```
Aprillion, line 104: on the main thread? can I see a deployed version, please? how is the performance on non-Chrome browsers, especially on mobile phones?

mruwnik: https://stampy-ui.ai688.workers.dev/ - firefox works fine, but mobile might be an issue

app/hooks/search.tsx
mruwnik: its not only used once. It's used every time someone searches for something with 1 or 2 words - https://github.com/StampyAI/stampy-ui/pull/260/files#diff-c7d7c7985c73cdd1f606a88adf3be2bf2c177d2045a2b22523ab502bf6d59218R162

Aprillion: the "lazy init" function in `useState(() => baselineSearch)` is only called on mount by React

app/hooks/search.tsx
Aprillion: I don't see any search results whatsoever in Firefox on apple-silicon Mac:

Screenshot 2023-06-08 at 10 47 00

Aprillion: they appeared when I checked the tab later, but it's not acceptable to not show anything for ~30 seconds or so

mruwnik:
Actually, it seems that it'll be worth doing anyway, for https://github.com/StampyAI/stampy-ui/issues/67 :/

app/hooks/search.tsx
```diff
@@ -0,0 +1,179 @@
...
+    if (!searchResults) return
+
+    if (queue[queue.length - 1] == userQuery) {
+      setResults(searchResults)
+    }
+    setQueue(updateQueue(userQuery))
+  }
+
+  // Each search query gets added to the queue of searched items - the idea
+  // is that only the last item is important - all other searches can be ignored.
```
mruwnik, line 161: this could be done with a simple variable to mark what the last search was, but a queue seems prettier? I'm not sure why

app/hooks/search.tsx
```diff
@@ -0,0 +1,179 @@
...
+    return []
+  }
+
+  const searchQueryTokens = normalize(searchQueryRaw).split(' ')
+  const matchers = searchQueryTokens.map((token) => ({
+    weight: token.match(/^(?:\w|\w\w|wh.*|how)$/) ? 0.2 : token.length,
+    fullRe: new RegExp(`\\b${token}\\b`),
+    prefixRe: new RegExp(`\\b${token}`),
+  }))
+  const totalWeight = matchers.reduce((acc, {weight}) => acc + weight, 0.1) // extra total to only approach 100%
```
Aprillion, line 50: 🤪apparently I'm terrible at comments - I totally forgot WHY I added that extra `0.1` total "to only approach 100%" ¯\\\_(ツ)_/¯

Aprillion (approved):
LGTM now 🤞

Issue #253, opened 2023-06-05
Deactivate default infinite scroll, and only activate on click


markovial:
As per the recent dev and editor meetings, we have decided on four main landing page questions to reduce the feeling of being overwhelmed when first navigating to the site. They are generally of the form:

- I am not convinced about AI Safety?
- I need more technical details about AI Safety?
- What are people working on in AI Safety?
- How can I help AI Safety?

Each of these have a fixed set of related questions underneath. So this means that infinite scroll should be by default off when someone first lands on the page. At the bottom of these questions there should be a button or some text that says something like - _Show me even more more questions about AI Safety..._ , or something like that, which when clicked turns on the infinite scroll generating random questions.

PR #261, opened 2023-06-07
optional infinisroll

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/253

image

There is still a slight potential issue where clicking on the stampy icon resets the questions, but not the infiniscroll. Dunno if that's good or bad, though

### DIFF
--- a/app/routes/index.tsx
+++ b/app/routes/index.tsx
@@ -1,6 +1,11 @@
-import {useEffect, MouseEvent, useRef} from 'react'
+import {useEffect, useState, useCallback, MouseEvent, useRef} from 'react'
 import type {LoaderFunction} from '@remix-run/cloudflare'
-import {ShouldRevalidateFunction, useOutletContext, useLoaderData} from '@remix-run/react'
+import {
+  ShouldRevalidateFunction,
+  useOutletContext,
+  useLoaderData,
+  useSearchParams,
+} from '@remix-run/react'
 import {loadInitialQuestions, QuestionState} from '~/server-utils/stampy'
 import {TOP} from '~/hooks/stateModifiers'
 import useQuestionStateInUrl from '~/hooks/useQuestionStateInUrl'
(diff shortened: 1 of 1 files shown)


app/routes/index.tsx
```diff
@@ -25,8 +30,72 @@ export const loader = async ({request}: Parameters<LoaderFunction>[0]) => {
   }
 }
 
+enum LoadMoreType {
```
mruwnik, line 33: possible ways of handling loading more questions

app/routes/index.tsx
```diff
@@ -25,8 +30,72 @@ export const loader = async ({request}: Parameters<LoaderFunction>[0]) => {
...
+  isSingleQuestion,
+  fetchMore,
+}: {
+  isSingleQuestion: boolean
+  fetchMore: () => Promise<any>
+}) => {
+  const [remixSearchParams] = useSearchParams()
+  const urlLoadType = useCallback(() => {
+    const more = remixSearchParams.get('more')
+    if (more) return LoadMoreType[remixSearchParams.get('more') as keyof typeof LoadMoreType]
```
mruwnik, line 52: the `more` url param can be used to manually specify what way of loading more questions should be used

app/routes/index.tsx
```diff
@@ -25,8 +30,72 @@ export const loader = async ({request}: Parameters<LoaderFunction>[0]) => {
...
+
+  useEffect(() => {
+    const more = urlLoadType()
+    if (more) {
+      setLoadMore(more)
+    }
+  }, [setLoadMore, urlLoadType])
+
+  const buttonHandler = () => {
+    fetchMore()
```
mruwnik, line 67: each button press should load more questions. This handles both basic "press for more questions" and switching to infiniscroll, as in the later case the infiniscroll might not be triggered right away on bigger screens

Aprillion (approved):

Issue #263, opened 2023-06-09
Google only indexes 5 questions


plexish:
https://www.google.com/search?q=site%3Aaisafety.info&oq=site%3Aaisafety.info

This would be really good to have fixed so we're more discoverable! @mruwnik?

mruwnik:
Looks like google has started to index it. It might still require all kinds of SEO black magic for the results to be higher

PR #264, opened 2023-06-09
Add robots.txt and sitemap

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/263


app/routes/[robots.txt].ts
```diff
@@ -0,0 +1,23 @@
...
+  // set up our text content that will be returned in the response
+  const origin = new URL(request.url).origin
+  const robotText = `
+User-agent: Googlebot
+Disallow: /nogooglebot/
+
+User-agent: *
+Allow: /
+
+Sitemap: ${origin}/sitemap.xml
```
Comment (mruwnik) [inline app/routes/[robots.txt].ts:14]: is there a better way to create this file? it's static other than this one url, which really only needs the origin

app/routes/[sitemap.xml].tsx
```diff
@@ -0,0 +1,27 @@
...
+  const formatQuestion = ({pageid, updatedAt}: Question) => `
+    <url>
+      <loc>${origin}/?state=${pageid}</loc>
+      <lastmod>${updatedAt}</lastmod>
+      <priority>1.0</priority>
+    </url>
+    `
+  const {data} = await loadAllQuestions(request)
+
+  const urls = data.filter((q) => q.status == QuestionStatus.LIVE_ON_SITE).map(formatQuestion)
```
Comment (mruwnik) [inline app/routes/[sitemap.xml].tsx:15]: I'm assuming that only live questions should be indexed...

app/routes/[robots.txt].ts
Comment (Aprillion) [reply app/routes/[robots.txt].ts:14]: I would say `https://aisafety.info/sitemap.xml` since we don't want to index any other domains 😅

Comment (mruwnik) [reply app/routes/[robots.txt].ts:14]: what about stampy.ai?

Comment (mruwnik) [reply app/routes/[robots.txt].ts]: Perfect! That makes things a lot simpler

Aprillion (approved):

Issue #222, opened 2023-04-29
Open outgoing links in a new tab


diziet:
All outgoing links on Stampy overwrite the current page, overwriting the existing browser tab. An alternative is to open new outgoing links in a new tab for the user by default.

There are pros and cons for both approaches. I wanted to list the major cons first:

Cons:
- Opening links in new tabs can often lead to confusion for users on mobile devices
- Removes the choice from the user (ie, a small fraction of users know to hold command when clicking or know to right click "open new tab"
- Can create confusion over internal links to Stampy SPA (single page app) links vs external links

Pros:
- As most outgoing links on the pages are large multi-page documents that might take 5-20 minutes for a fast reader to go through, opening links in a new tab creates a potential reading list while keeping the current navigated question open.
Ie, on the [first](https://aisafety.info/?state=8486_6714r6178r8428r6194-6708-6172-6479-6968-881W-8163-6953-5635-8165-7794-6715-) page on Stampy, the first 4 links are all very long articles.
- Selfishly retain the hierarchical top level stampy navigation in the user's browser for as long as possible while also making a space for the high quality outgoing links, increasing the chance the user engages with the content later

plexish:
Has been requested by the distillation fellows, I'm also in favor of it, but I think @Aprillion was not sure? Overall seems the right choice.

Aprillion:
distillation fellows should be able to open links in a new tab from any webpage, regardless of whether stampy will mess up with web standards or not

my own dev experience also doesn't matter

please make a UX research with a sample of people who are the most important intended audience, thanks! some UX folks wanted to help, right?

martinkunev:
To counter the following:
"Removes the choice from the user (ie, a small fraction of users know to hold command when clicking or know to right click "open new tab""

Consider two categories of users:
* those who can make the choice - These users know how to navigate between tabs. Opening external links in new tabs is more or less the standard behavior so I think those users would hardly be surprised.
* those who don't know / don't care - Such a user could be equally surprised by disappearing pages and by new tabs. Disappearing pages is less desirable because it leads to information loss.

Aprillion:
When the next LLM is trained on this publicly available example of how a group of people who care about AGI Alignment behaved, what do we want it to discover, what values should it simulate in a situation when we don't know what actual users want?

That we made some reasonable assumptions and made a decision ourself or that we asked a sample of the users for their actual opinion?

robertskmiles:
[...]

PR #269, opened 2023-07-05
Open external links on new tab

Aprillion:
Code cleanup & fixing #222 

image

Testable on https://stampy-ui.aprillion.workers.dev/


Aprillion:
aaah, we don't use post css and Chrome only supports `-webkit-mask`, need up update 😓

mruwnik (approved):
Approved, because the code seems correct, and clicking on the links works.
At the same time, I'd like to note that I agree with your initial doubts, and so am disappointed at how you didn't wait for someone else to implement it, hence failed condemning it to languish in the backlog :P

ccstan99:
Sorry to be the pain but my fault for reviving the languishing zombie. Unfortunately, it's the right thing to do from a user's expectation point of view and for metrics, although it'll open a different set of issues for editors to rethink how they structure content and links.

Aprillion:
This is part of our code that wouldn't make a good first issue to welcome new developers, and it was deemed important by many stakeholders in our team, so I decided to implement it so the issue stops haunting us 🤷‍♂️ 

@mruwnik feel free to add a condition like `if (!localStorage.get('disableExternalLinksOpenOnNewTab'))` if you think it might be important to have that option (and we could ask to include it as a user-visible setting in the UI redesign later this year)

ccstan99:
@Aprillion @mruwnik Just so I better understand your concerns... Is the problem nasty code maintenance? Or users overwhelmed with too many new tabs? If it's the latter, we could ask the user which they prefer ("Open in new tab or replace current one?" with an option "Remember this setting." Or does require cookies that we need to avoid?

mruwnik:
no, it's more about changing standards. Or leaving things as default as possible. It's part of the general process of websites taking on more and more of the basic functionality of browsers, which can be helpful, but often just makes things more counter intuitive. Though this may be a matter of "get off my lawn!!"? :D

It takes away freedom in the name of temporary comfort. And breaks one of the potential user flows.

I don't think having millions of options like this would help much. That's going back to badly reimplementing the browser in JS.

PR #282, opened 2023-07-18
load all tags before questions

mruwnik:


app/routes/index.tsx
```diff
@@ -22,6 +22,7 @@ import {reloadInBackgroundIfNeeded} from '~/server-utils/kv-cache'
 
 export const loader = async ({request}: Parameters<LoaderFunction>[0]) => {
   try {
+    await loadTags(request)
```
mruwnik, line 25: double the initial loading time :/
The problem here is that questions don't download the metadata of their tags, only the name. So the questions need to fetch tags before being able to filter out internal ones. This adds an initial tax, rather than per request, so potentially better?

Aprillion: did you check `/questions/cache` on localhost that it's cached, please?

mruwnik: it is

Aprillion (approved):

Issue #295, opened 2023-08-01
/random -> loads one random on site answer


plexish:
Rob requested a feature where https://aisafety.info/random loads a random question. This will help him do quality review, and would likely be used by fellows too. It does not need to appear on the homepage, can be hidden feature.

PR #296, opened 2023-08-03
random questions

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/295

### DIFF
--- a/app/routes/random.tsx
+++ b/app/routes/random.tsx
@@ -1,0 +1,60 @@
+import {useEffect} from 'react'
+import type {LoaderFunction} from '@remix-run/cloudflare'
+import {useOutletContext, useLoaderData} from '@remix-run/react'
+import {reloadInBackgroundIfNeeded} from '~/server-utils/kv-cache'
+import {loadAllQuestions, QuestionState, QuestionStatus} from '~/server-utils/stampy'
+import {Header, Footer} from '~/components/layouts'
+import {Question} from '~/routes/questions/$question'
+import useQuestionStateInUrl from '~/hooks/useQuestionStateInUrl'
+
+function randomItem(arr: any[]) {
+  return arr[Math.floor(Math.random() * arr.length)]
+}
+
+export const loader = async ({request}: Parameters<LoaderFunction>[0]) => {
+  try {
+    const {data} = await loadAllQuestions(request)
+    return {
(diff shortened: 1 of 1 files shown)


mruwnik:
ping

ccstan99 (approved):
LGTM! Don't have time to test it out but I know Rob/plex want this as a priority. @Aprillion feel free to add anything else you feel relevant.

app/routes/random.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export const loader = async ({request}: Parameters<LoaderFunction>[0]) => {
+  try {
+    const {data} = await loadAllQuestions(request)
+    return {
+      ...randomItem(data.filter((q) => q.status == QuestionStatus.LIVE_ON_SITE)),
+      questionState: QuestionState.OPEN,
+    }
+  } catch (e) {
+    console.error(e)
```
Aprillion, line 22: what's the deal with the server side logging here, please? does it do anything different than letting the exception unhandled?

why not send the error to client a la https://github.com/StampyAI/stampy-ui/blob/master/app/routes/questions/%24question.tsx ?

app/routes/random.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const question = useLoaderData<ReturnType<typeof loader>>()
+
+  return (
+    <>
+      <Header />
+      <main>
+        <Question
+          questionProps={question}
+          onLazyLoadQuestion={onLazyLoadQuestion}
+          onToggle={toggleQuestion}
```
Aprillion, line 51: toggle does not work for me

Aprillion (commented):
looks good for initial implementation

FTR there is usually no point pinging me on the same platform, I'm not checking emails that much during vacation 😅 I would have noticed https://discord.com/channels/677546901339504640/1125882087908573305 sooner if needed next time (though also not 100%)

app/routes/random.tsx
mruwnik: doesn't work for me, either. Though it should be fine? It would be nice for this to also handle related questions and the search bar, but it didn't seem worth the bother...

Aprillion: would be even better if there was no `>` chevron if the feature doesn't make sense .. which it probably doesn't 😅

a refresh button at the bottom (similar to load more questions) saying `Reload for a another random question` might also be nice some time in the future

PR #297, opened 2023-08-04
fix glossary bugs

mruwnik:
https://github.com/StampyAI/GDocsRelatedThings/issues/54
https://github.com/StampyAI/GDocsRelatedThings/issues/55
https://github.com/StampyAI/GDocsRelatedThings/issues/56

### DIFF
--- a/app/routes/questions/$question.tsx
+++ b/app/routes/questions/$question.tsx
@@ -173,7 +173,7 @@ export function Question({
 const updateTextNodes = (el: Node, textProcessor: (node: Node) => Node) => {
   Array.from(el.childNodes).forEach((child) => updateTextNodes(child, textProcessor))
 
-  if (el.nodeType == Node.TEXT_NODE && el.textContent && el?.parentElement?.tagName != 'A') {
+  if (el.nodeType == Node.TEXT_NODE && el.textContent) {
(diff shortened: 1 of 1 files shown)


app/routes/questions/$question.tsx
```diff
@@ -173,7 +173,7 @@ export function Question({
 const updateTextNodes = (el: Node, textProcessor: (node: Node) => Node) => {
   Array.from(el.childNodes).forEach((child) => updateTextNodes(child, textProcessor))
 
-  if (el.nodeType == Node.TEXT_NODE && el.textContent && el?.parentElement?.tagName != 'A') {
```
Aprillion, line 176: please check git commit for the old code, it was fixing (workarounding?) some previous bug - not sure which, but it deserves to be re-tested that removing the last condition won't reintroduce the previous bug 🤞

Aprillion (approved):
other than the potential regression, LGTM 👍

app/routes/questions/$question.tsx
mruwnik: it was checking whether the glossary item is part of a link, so it didn't display links as glossary items. But apparently the editors want that... :D

Issue #294, opened 2023-08-01
Tags disappearing and reappearing on-site


ChatrikMangat:
Tags seem to disappear and reappear on-site for some questions. I've included an example in the images below. This is the same [state](https://aisafety.info/?state=6306_6586r6957r8IHOr6966r1001_9B85r6410r6411r6412r7602r8NYDr6196r8E40r5864_6207r2374r6587r5633r8158r9OGZ-9IDQ-9TDI-8TJV-) captured around 2 hours apart. Two other questions opened in the state "What are the differences between AGI, transformative AI, and superintelligence?" and "What is an intelligence explosion?" also show similar behavior, not in sync though. 

As I type this out, only the example included in the images below has visible tags on-site, the other two questions don't, although they have fluctuated in the past few days.

Tags disappeared:
![tags2_21_44](https://github.com/StampyAI/stampy-ui/assets/50947666/8484eb92-ecf2-42ac-9d35-9fa4a45fe5af)

Tags reappeared around 2 hours later: 
![tags2_23_39](https://github.com/StampyAI/stampy-ui/assets/50947666/969785fd-af5a-4614-a3a7-8d2e013f5013)

Coda table entry for the post:
![tags2_coda](https://github.com/StampyAI/stampy-ui/assets/50947666/13521d74-0ca1-4b00-8dee-573982cb7922)

PR #298, opened 2023-08-05
Make sure the tags are loaded when fetching questions contents

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/294

### DIFF
--- a/app/server-utils/stampy.ts
+++ b/app/server-utils/stampy.ts
@@ -258,6 +258,11 @@ const convertToQuestion = ({name, values, updatedAt} = {} as AnswersRow): Questi
 })
 
 export const loadQuestionDetail = withCache('questionDetail', async (question: string) => {
+  // Make sure all tags are loaded. This shouldn't be needed often, as it's double cached
+  if (Object.keys(allTags).length === 0) {
+    const {data} = await loadTags('NEVER_RELOAD')
+    allTags = Object.fromEntries(data.map((r) => [r.name, r])) as Record<string, Tag>
+  }
   const rows = (await getCodaRows(
     QUESTION_DETAILS_TABLE,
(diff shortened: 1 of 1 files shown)


app/server-utils/stampy.ts
```diff
@@ -258,6 +258,11 @@ const convertToQuestion = ({name, values, updatedAt} = {} as AnswersRow): Questi
 })
 
 export const loadQuestionDetail = withCache('questionDetail', async (question: string) => {
+  // Make sure all tags are loaded. This shouldn't be needed often, as it's double cached
+  if (Object.keys(allTags).length === 0) {
+    const {data} = await loadTags('NEVER_RELOAD')
+    allTags = Object.fromEntries(data.map((r) => [r.name, r])) as Record<string, Tag>
```
mruwnik, line 264: I'm not sure whether this is the actual cause. Though I can't see anything else that might be an issue. It's a pity that Coda doesn't have proper joins, which would solve all of these issues...

My suspicion is that cloudflare changes workers, which means that the first one works fine, as it set `allTags` to contain the tags, but then:
* it calls `loadTags()` in index.tsx
* the cached value is returned (so it didn't actually call the function)
* `allTags == {}` for that worker

This will protect from that, at the potential cost of more calls

Aprillion: I'm slightly worried about long term maintainability if we get into the habit of optimistic fixing / making workarounds without understanding root cause 🤔 

but currently on vacation and I don't see anything obviously bad about this PR itself, so YOLO approving 🤷

Aprillion (approved):

app/server-utils/stampy.ts
mruwnik: agreed :/

Issue #290, opened 2023-07-31
Export questions + relations to twine


markovial:
It would be useful to have a visual representation of the question relations, as well as the ability to edit them. So export the list of all questions + relateds in the google doc to [twine](https://twinery.org/).

I don't know if the export should be triggered manually, or should just happen every once in a while? If there have been changes made in twine which are not reflected on the google doc, then exporting might overwrite those changes...?

Right now the plan is to only export, and manually do a diff and then make the appropriate changes in google doc.

Futher down the line, there should probably also be a button which on click manually imports back the diff'ed changes to the relateds? I don't know if we want to import back automatically periodically because we still havent decided on the source of truth yet.

PR #299, opened 2023-08-06
Add page for editors

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/290

This collects the hidden functionality in a single, also hidden functionality. I'm not at all sure whether this is a good idea, though...

image


Aprillion (approved):
yeah, this is much simpler than having to log in and seeing different buttons on each question and other editor features, good low hanging fruit to hopefully improve editor experience a bit 👍

Issue #292, opened 2023-07-31
Feature Request: Wikipedia style banners


markovial:
It would be nice to have different banners for live on site content based on 'coda tags' ("needs work", "technical", "work in progress" etc.). So if something needs work that shows up as a banner under the question.

I think something like [this on wikipedia](https://en.wikipedia.org/wiki/Category:WikiProject_banner_templates) is what Rob had in mind, but I'm not sure.

The banners will pull from a "tag" which will be applied to that question on coda. Either manually or automagically when the question is being edited or something.

This is the github issue for creating the banner not the coda table.

PR #300, opened 2023-08-06
Add basic banner functionality

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/292

image

This is just adds the basic functionality, without any decent colours or images (both of which can be set in Coda)

### DIFF
--- a/app/routes/questions/$question.tsx
+++ b/app/routes/questions/$question.tsx
@@ -1,13 +1,8 @@
 import type {LoaderArgs} from '@remix-run/cloudflare'
-import {
-  GlossaryEntry,
-  loadQuestionDetail,
-  QuestionState,
-  QuestionStatus,
-} from '~/server-utils/stampy'
+import {GlossaryEntry, loadQuestionDetail, QuestionState} from '~/server-utils/stampy'
 import {useRef, useEffect, useState} from 'react'
 import AutoHeight from 'react-auto-height'
-import type {Question, Glossary, PageId} from '~/server-utils/stampy'
+import type {Question, Glossary, PageId, Banner as BannerType} from '~/server-utils/stampy'
 import type useQuestionStateInUrl from '~/hooks/useQuestionStateInUrl'
 import {Edit, Link as LinkIcon} from '~/components/icons-generated'
 import {Tags} from '~/routes/tags/$tag'
@@ -34,6 +29,7 @@ export const loader = async ({request, params}: LoaderArgs) => {
       answerEditLink: null,
       relatedQuestions: [],
       tags: [],
+      banners: [],
     }
     return {
       error: error?.toString(),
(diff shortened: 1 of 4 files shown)


app/routes/questions/$question.tsx
```diff
@@ -166,6 +169,29 @@ export function Question({
...
+      }}
+    >
+      <h3>
+        <img src={icon?.url} alt={icon?.name} />
+        <span className="title">{title}</span>
+      </h3>
+      <div
+        className="banner-contents"
+        dangerouslySetInnerHTML={{
+          __html: text,
```
mruwnik, line 188: or should it only allow basic text, rather than properly handling markdown?

app/server-utils/stampy.ts
```diff
@@ -263,6 +288,10 @@ export const loadQuestionDetail = withCache('questionDetail', async (question: s
     const {data} = await loadTags('NEVER_RELOAD')
     allTags = Object.fromEntries(data.map((r) => [r.name, r])) as Record<string, Tag>
   }
+  if (Object.keys(allBanners).length === 0) {
```
mruwnik, line 291: isn't this fun...

app/routes/questions/$question.tsx
```diff
@@ -70,17 +66,7 @@ export function Question({
...
-    pageid,
-    title: codaTitle,
-    status: codaStatus,
-    text,
-    answerEditLink,
-    questionState,
-    tags,
-  } = questionProps
-  const title =
-    codaStatus && codaStatus !== QuestionStatus.LIVE_ON_SITE ? `WIP - ${codaTitle}` : codaTitle
```
mruwnik, line 83: a banner can be displayed for non published questions, rather than the `WIP - `

app/server-utils/stampy.ts
Aprillion: feel free to create an issue where you collect all notes about what needs to be double checked and loaded in correct order currently and deserves simplification, I might get to it during aisafety.com hackathon 🤞

app/server-utils/stampy.ts
```diff
@@ -263,6 +288,10 @@ export const loadQuestionDetail = withCache('questionDetail', async (question: s
...
+    const {data} = await loadBanners('NEVER_RELOAD')
```
Aprillion, line 292: this is the only place where it's called? `NEVER_RELOAD` doesn't sound right, normal cache should be used with reloading in background after 10 minutes IMHO 🤔

app/routes/questions/$question.tsx
mruwnik: they all have the appropriate banner set, so it's mainly a matter of deciding whether that's the way to go. The parser is already updated to do so

Aprillion (approved):

app/routes/questions/$question.tsx
mruwnik: oh, this most certainly is a bad choice - I pretty much just plopped random colours there to "encourage" someone else to choose decent ones :D

Issue #277, opened 2023-07-10
Feature request: Searching Stampy directly from a browser's search bar


LeMurphant:
It would be nice to have a url for direct search of aisafety.info in my browser's search bar with something like this
https://arxiv.org/search?query=%s

Ccstan pointed out that this works, but the UI isn't great and links are not clickable
https://nlp.stampy.ai/api/search?query=agi

I believe it would mostly be used by content writers, but could be a nice touch.

mruwnik:
can already be done, e.g. [aisafety.info/?q=why not turn it off](https://aisafety.info/?q=why not turn it off)

LeMurphant:
@mruwnik Seems to work quite well in production. I tried it for the QA interface at https://nlp.stampy.ai/ and it did not seem to work, is this functionality something that can reasonably easily be ported to the QA interface to allow searching for in-progress articles?

mruwnik:
would you prefer for it to work on nlp.stampy.ai, or for aisafety.info to have a flag that allows searching for in progress articles? Or both? :D

LeMurphant:
Either would do. If we do it on aisafety.info we just want to make sure people don't stumble on it inadvertently

LeMurphant:
@mruwnik this seems broken with the new site. Then again, without something like https://github.com/StampyAI/stampy-ui/issues/571 I'm not sure how you would fix it as I'm not sure what page https://aisafety.info/?q=existential should return. How about the first result by default as a quick fix?

mruwnik:
first result with the search bar prepopulated and results shown, or just the first result?

LeMurphant:
I'd weakly favour the former if it's something that can be done while keeping navigation smooth.

LeMurphant:
Reopening as bryce asked since this is still relevant

Issue #254, opened 2023-06-05
Feature Request: LMGTFY Functionality


robertskmiles:
It would be cool if you could send someone a link that would pre-populate the search bar with a string. I imagine this could be done with a URL variable? Like: `https://aisafety.info/?query="How can I help?"`

robertskmiles:
For creating links, I'm not sure. Maybe a mousover link icon next to the search bar? I don't mind if it's a hidden feature, I just want to use this myself :)

As for how it looks, I'm imagining it just runs the search as normal and shows the drop-down as though the user had just typed in the search term. I'm not sure we need to make any new UI for this feature. If no question IDs are in the link, populate with the initial questions, otherwise we can load in the questions from the URL, with the search from the URL as well.

mruwnik:
how about something like `https://aisafety.info/question/H`https://aisafety.info/question/Howow%20can%20I%20help%32` (those aren't proper encodings, I'm just lazy) which redirects to `https://aisafety.info/?state=1234_`?

PR #303, opened 2023-08-20
Extract search

mruwnik:
* Refactor the search module so it's independent (no external dependancies)
* Extract the search module as a separate library
* Add an example server+page to showcase that the independent search works by itself
* Change the fronted to use the search library
* Add CORS for selected routes (including static resources)
* Allow [searching](https://github.com/StampyAI/stampy-ui/issues/254) directly [from the url](https://github.com/StampyAI/stampy-ui/issues/277)

This can also be seen in action [here](https://carmelita561.preview.softr.app/tester?show-toolbar=true) - might need to press enter for it to start searching. This involves hacky malusage of HTML, so is very clunky - I couldn't be bothered to do it properly.

A deployed version of this can be found [here](https://stampy-ui.ai688.workers.dev/?query=will+this+work)


stampy-search/README.md
Aprillion, line 1: This PR is too big for me to review right now, should get to it around August 30...

I noticed the deployed version does not update the search query in URL when updating search box.

Aprillion: URL is not getting any shorter by not updating it:
Screenshot 2023-08-22 at 14 22 21

Feel free to completely remove the URL state if people don't need it. But if it's needed, it should be correct.

app/components/search.tsx
```diff
@@ -1,36 +1,69 @@
...
+      if (value === searchInputRef.current) return
 
-    searchInputRef.current = value
+      searchInputRef.current = value
 
-    search(value)
-    logSearch(value)
-  }
+      setPendingSearches(true)
+      searcher && searcher.searchLive(value)
```
Aprillion, line 55: this is probably mostly for TS reasons, but it shouldn't be a valid option to NOT perform the search at this step, if it "always works" it should throw an exception after future changes that make it suddenly "not work"

app/components/search.tsx
```diff
@@ -1,36 +1,69 @@
...
 
-  const {search, arePendingSearches, results} = useSearch(onSiteAnswersRef)
+  const [arePendingSearches, setPendingSearches] = useState(false)
+  const [results, setResults] = useState([] as SearchResult[])
+  const [searcher, setSearcher] = useState<Searcher>()
+
+  useEffect(() => {
+    setSearcher(
+      new Searcher({
+        getAllQuestions: () => onSiteAnswersRef.current,
```
Aprillion, line 36: This looks like a very useful exercise to find out how reusable is current architecture, and I'd be happy for this code to live on a branch or fork for experimental deployments.

However, I am not able to confirm this is the right step for the master branch and production deployment, where we should aim for more maintainability and simplified architecture. Adding more features on top of previous hacks does not sound sustainable to me.

For 3rd party users of this feature, creating new class instances to spin of workers sensitive to CORS and special API calls while having to provide your own `getAllQuestions`, then dealing with optional-results success callbacks without error handling and quite complex semantics ... that doesn't sound very simple to use / modular / ...

If there is someone else who wants to take responsibility for long term maintenance of this repo with occational firefighting and debugging if the site stops working, I'd be happy to hand over the "architecture oversight".

If not, I will rather focus on figuring out how to include stampy search on 3rd party sites using some simpler API, without this new PoC feature 🤔 Ideally something drop-in replacable with API from https://nlp.stampy.ai/ (or whatever is / will be the current url / name for that feature)

Aprillion:
We are doing #307 at the aisafety.com hackathon and we found a compromise between practicality, magic, and at least halfway hoping it's still maintainable in PRs related to that issue => let's keep this branch for inspiration of future improvements, but not merging to master so closing the PR.

PR #305, opened 2023-08-23
Add CORS to /questions/$

mruwnik:
The chatbot is playing about at proxying this, when it could just directly call it

### DIFF
--- a/app/server-utils/responses.ts
+++ b/app/server-utils/responses.ts
@@ -1,4 +1,4 @@
-import {json} from '@remix-run/cloudflare'
+import {LoaderArgs, json} from '@remix-run/cloudflare'
 
 export const jsonCORS = <T>(data: T) =>
   json(data, {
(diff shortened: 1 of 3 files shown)


app/server-utils/responses.ts
```diff
@@ -7,3 +7,29 @@ export const jsonCORS = <T>(data: T) =>
...
+        'Access-Control-Allow-Headers': 'Content-Type, Authorization, allow-control-allow-origin',
+      },
+    }
+  )
+}
+
+type Handler = (args: LoaderArgs) => any
+export const wrapCORS = (fn: Handler) => async (args: LoaderArgs) => {
+  if (args.request.method == 'OPTIONS') {
+    return CORSOptions(args.request)
```
Aprillion, line 31: is any of this special to cloudflare or remix? if yes, a link to relevant documentation would be nice ... if not, if all just standard HTTP server response stuff 👍

Aprillion (approved):

app/server-utils/responses.ts
mruwnik: no. This is to get the chat bot working, which uses remix. This is all bog standard http stuff

Issue #307, opened 2023-09-02
Unbranded version for embedding via iframe


Aprillion:
Make a version of the UI for other websites (specifically, aisafety.com) to embed the stampy search bar from aisafety.info as an `<iframe>`.

Include most features but without header, footer, and the homepage default list: search bar + results + "something else" button and modal + display answer in the iframe (assuming hosting site can implement dynamic height resizing). Option to link to aisafety.info answer pages instead of showing directly in iframe is nice to have.

Make an example HTML page with `<iframe>`, for copy&pasting to tools that support HTML but not npm packages, like [Soft](https://studio.softr.io/) (used for aisafety.com).

Aprillion:
* [ ] dark mode as url parameter (so that hosting site can control it)
* [ ] embedded as route or url parameter
* [ ] remove logo, header, footer, initial list
* [ ] make example iframe route
* [ ] dynamic height of iframe
* [ ] include on aisafety.com

Aprillion:
We figured out a way to set dynamic height of the iframe by sending *window.parent.[postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage)* from the embedded page and updated https://aisafety.info/embed-example.html to demonstrate how to do it on parent page.

PR #308, opened 2023-09-02
Version without header/footer for embedding in iframes

Aprillion:
Hopefully MPV quality for #307
 => let me know if anyone spots bugs: https://stampy-ui.aprillion.workers.dev/embed-example.html

Screenshot 2023-09-02 at 19 51 14


mruwnik (approved):
I don't like how this is another layer of hackiness, but as long as it works...

PR #309, opened 2023-09-03
Make links out of search results and question titles when embedded

Aprillion:
Improvements for #307 => testing on https://aisafety.preview.softr.app/new
* can be controlled by `showDetails` url param
* I was not able to reproduce dynamic height in Softr, either CSP or it doesn't allow listening to events somehow, so this is a version more friendly for static height

### DIFF
--- a/app/hooks/useQuestionStateInUrl.ts
+++ b/app/hooks/useQuestionStateInUrl.ts
@@ -36,6 +36,8 @@ const emptyQuestionArray: Question[] = []
 export default function useQuestionStateInUrl(minLogo: boolean, initialQuestions: Question[]) {
   const [remixSearchParams] = useSearchParams()
   const transition = useTransition()
+  const embedWithoutDetails =
+    remixSearchParams.has('embed') && !remixSearchParams.has('showDetails')
 
   const [stateString, setStateString] = useState(
     () =>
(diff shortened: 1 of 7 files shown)


app/hooks/useQuestionStateInUrl.ts
```diff
@@ -194,6 +196,12 @@ export default function useQuestionStateInUrl(minLogo: boolean, initialQuestions
   const toggleQuestion = useCallback(
     (questionProps: Question, options?: {moveToTop?: boolean; onlyRelated?: boolean}) => {
       const {pageid, relatedQuestions} = questionProps
+
+      if (embedWithoutDetails) {
+        window.open(`https://aisafety.info/?state=${pageid}_`, '_blank')
```
mruwnik, line 201: I don't like how this is getting even more magical. Would it be possible to do this somewhere closer to the actual click?

mruwnik (approved):

app/hooks/useQuestionStateInUrl.ts
Aprillion: yeah ... this was the shared code between all places that open question details, so it's kinda DRY 🏝️ 

we should probably render real `<a>` links in the version of page without details, but if we do that in current components it would be more code that is still not self-explanatory and half-magic, even if not as much as this 🤔 I will try to make that version and see which will be less horrible 😅 

the real solution would be to take the old code as inspiration and do the React components from scratch with all the existing use cases in mind, but I don't have time for that on this hackathon and not really the highest priority for the site either I guess 😭

Aprillion: the refactored version to use real links: #310

mruwnik (approved):

PR #311, opened 2023-09-03
Support shareable url with search query (q and limit)

Aprillion:
Replacing feature from #303 to be able to share url to search results instead of just directly to answer.

Some fixes for URL state when overwriting search and using the browser back/forward buttons, seems good enough for this hidden feature, but probably not 100% bugfree yet 😅 

Testable in https://stampy-ui.aprillion.workers.dev/?q=why&limit=100

### DIFF
--- a/app/components/search.tsx
+++ b/app/components/search.tsx
@@ -1,4 +1,4 @@
-import {useState, useEffect, useRef, MutableRefObject, FocusEvent} from 'react'
+import {useState, useEffect, useRef, MutableRefObject, FocusEvent, useCallback} from 'react'
 import debounce from 'lodash/debounce'
 import {AddQuestion} from '~/routes/questions/add'
 import {Action, ActionType} from '~/routes/questions/actions'
@@ -38,7 +38,7 @@ export default function Search({
 
   const {search, arePendingSearches, results} = useSearch(onSiteAnswersRef, limitFromUrl)
 
-  const searchFn = (rawValue: string, initialSearch = false) => {
+  const searchFn = (rawValue: string) => {
     const value = rawValue.trim()
     if (value === searchInputRef.current) return
 
@@ -46,15 +46,30 @@ export default function Search({
 
     search(value)
     logSearch(value)
+  }
 
-    if (queryFromUrl && !initialSearch) {
-      removeQueryFromUrl()
+  // run search if queryFromUrl is provided initially or if it pops from browser history after it was removed,
+  // update url if searchInput changes,
+  // and use current version of functions without affecting deps
+  const searchInput = searchInputRef.current
+  const searchFnRef = useRef(searchFn)
+  searchFnRef.current = searchFn
+  const removeQueryFromUrlRef = useRef(removeQueryFromUrl)
+  removeQueryFromUrlRef.current = removeQueryFromUrl
+  const queryFromUrlWasRemoved = useRef(false)
+  useEffect(() => {
+    if (queryFromUrl) {
+      if (!searchInput || queryFromUrlWasRemoved.current) {
+        searchFnRef.current(queryFromUrl)
+        queryFromUrlWasRemoved.current = false
+        const inputEl = document.querySelector('input[name="searchbar"]') as HTMLInputElement
+        inputEl.value = queryFromUrl
+      } else if (queryFromUrl !== searchInput) {
+        removeQueryFromUrlRef.current()
+        queryFromUrlWasRemoved.current = true
+      }
     }
-  }
-
-  if (queryFromUrl && !searchInputRef.current) {
-    searchFn(queryFromUrl, true)
-  }
+  }, [queryFromUrl, searchInput])
 
   const handleChange = debounce(searchFn, 100)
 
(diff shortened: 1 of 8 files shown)


app/routes/index.tsx
```diff
@@ -24,8 +24,10 @@ import type {Context} from '~/root'
 const empty: Awaited<ReturnType<typeof loadInitialQuestions>> = {data: [], timestamp: ''}
 export const loader = async ({request}: Parameters<LoaderFunction>[0]) => {
   const embed = !!request.url.match(/embed/)
+  const queryFromUrl = !!request.url.match(/[?&]q=/)
   const showInitial = !!request.url.match(/showInitial/)
-  if (embed && !showInitial) return {initialQuestionsData: empty}
+  const hideInitial = (embed || queryFromUrl) && !showInitial
```
mruwnik, line 29: [NIT] maybe `fetchInitial` and `if(!fetchInitial)`?

app/routes/index.tsx
```diff
@@ -220,18 +227,17 @@ export default function App() {
...
-            isSingleQuestion={
-              questions.filter((i) => i.questionState != QuestionState.RELATED).length == 1
-            }
-          />
-        </>
+        <a id="discordChatBtn" href="https://discord.com/invite/Bt8PaRTDQC">
+          <Discord />
+        </a>
+      )}
+      {!embed && !queryFromUrl && (
```
mruwnik, line 234: does this mean that querying in the url will disable the "fetch more" button? Why?

app/hooks/useQuestionStateInUrl.ts
```diff
@@ -158,7 +160,15 @@ export default function useQuestionStateInUrl(minLogo: boolean, initialQuestions
       if (stateString == newState) return
       const newSearchParams = new URLSearchParams(remixSearchParams)
       newSearchParams.set('state', newState)
-      history.replaceState(newState, '', '?' + newSearchParams.toString())
+
+      const newUrl = '?' + newSearchParams.toString()
+      if (allowBrowserBackToInitialStateRef.current) {
```
mruwnik, line 165: what is this for?

app/components/search.tsx
```diff
@@ -22,24 +25,35 @@ export default function Search({
...
+  const searchFn = (rawValue: string, initialSearch = false) => {
     const value = rawValue.trim()
     if (value === searchInputRef.current) return
 
     searchInputRef.current = value
 
     search(value)
     logSearch(value)
+
+    if (queryFromUrl && !initialSearch) {
```
mruwnik, line 50: I don't like this... How about wrapping the `  if (queryFromUrl && !searchInputRef.current) {` in a `useEffect` or something? That should allow ` initialSearch ` to be removed?

app/routes/index.tsx
Aprillion: I'm fine either way => changed

app/routes/index.tsx
Aprillion: The page without button looks nice and clean, but the extra button was very distracting for me, personally 😅 I can bring it back if more peple will want it, but would prefer to keep it out for now:

Screenshot 2023-09-04 at 10 49 48

app/hooks/useQuestionStateInUrl.ts
Aprillion: (will answer in next comment)

app/components/search.tsx
Aprillion: yeah, I would have to wrap some functions in a callback which would cascade to more functions in other hooks, but that pattern sounds easier to understand => will do

app/hooks/useQuestionStateInUrl.ts
Aprillion: sneaking in a fix for situation when someone starts at homepage or some other page from direct link, then opens a bunch of stuff and clicks browser back button hoping to land on the initial page ... but currently, we only replace state so back button takes you to previous domain and then forward takes you to final state not initial state, page refresh also keeps you on final state ... I wanted to be able to go back to initial page + combined with the logo reloading the page without "smart" consolidation of the state will create a more rich browser history so people can return to each restart in their journey (if someone wants to do multiple restarts by clicking on the logo - probably very niche use case)

so not polluting the browser history with every state change while remembering both the initial and the final state sounded like a good compromise to me 🤞

app/routes/index.tsx
mruwnik: you can get that by setting `more=disabled`. I'm fine either way

mruwnik (approved):
The usual caveats apply :D

app/components/search.tsx
Aprillion: now updated + 1 more fix for browser back button when using `?q=...` initially, then running manual search, then going back

app/routes/index.tsx
Aprillion: good point, we want the footer when it's just `queryFromUrl`, but no footer when it's `embed` 😅 so I ended up with a combination of approaches to cover both use cases

code updated

PR #313, opened 2023-09-04
iframe dynamic height using postMessage

Aprillion:
fix for dynamic height, testing on https://aisafety.softr.app/new-stampy

(advice from ChatGPT 4 helped 😅)


mruwnik (approved):
I don't know what this is doing, but it looks sane, the site seems to work, so approving :D

PR #314, opened 2023-09-09
onlyInitial url parameter for iframe embeds

Aprillion:
* also run prettier on html file and fix eslint warnings
* https://stampy-ui.aprillion.workers.dev/embed-example.html, now showing source code for multiple combos
image


README.md
```diff
@@ -11,10 +11,11 @@ Contributions are welcome, the code is released under the MIT License. If you'd
 - state - controls which questions are displayed as collapsed / open / related, e.g. [aisafety.info/?state=6568\_](https://aisafety.info/?state=6568_)
 - q (string) - search query for sharing direct link to search results (and not just link to 1 question), e.g. [aisafety.info/?q=alignment&limit=10](https://aisafety.info/?q=alignment&limit=10)
   - limit (number, default 5) - how many results to show
-- embed - show site without header/footer for embedding on other sites, se [embed-example.html](/public/embed-example.html)
+- embed - show site without header/footer for embedding on other sites, see [embed-example.html](https://aisafety.info/embed-example.html)
   - placeholder (string) - override `<input placeholder=...>` of the search box
   - theme (light|dark) - override CSS theme (if not provided, the embedded site will use `preferred-color-scheme` system setting)
-  - showInitial - also show initial questions, not just search bar
+  - showInitial - show initial questions as well as the search bar
```
mruwnik, line 17: how about something like `full`, `all` or even `default` for it to work like the main page?

mruwnik (approved):

README.md
Aprillion: not sure what you mean exactly, if it's about ininite scroll, the iframed page has overflow: hidden to control the iframe height of the parent page, so scroll-triggered loading wouldn't work in embedded version .. the button version of loading more should be possible to work - feel free to investigate and fix as needed to make it better 👍

mruwnik: ah, cool. I was under the impression that this `showIniitial` would just cause the iframe to behave like the real site

PR #317, opened 2023-09-15
Iframe with link fix

Aprillion:
* it didn't work after merge (because `handleSpecialLinks` has a condition `url.hostname === window.location.hostname` that was false on localhost and test envornment)
* so I reverted the merge
* and now redoing the changes + fixing the issue


Aprillion:
The deployed version after revert merge is still broken 🤔  merging this PR now and if there are left-over bugs we can solve later 🍀

Issue #321, opened 2023-09-22
Add alternative phrasings to basic search


mruwnik:
From discussion in https://discord.com/channels/677546901339504640/1125883308979535922/1154377481575485540 :

> Are alternate phrasings supposed to show up on the 1-2 word keyword search or only on semantic search?
> ...
> If it's an easy fix it might be worth including both

ccstan99:
This issue should only apply to keyword search. Both semantic search through in-browser TF and API already support alternate phrasings.

PR #326, opened 2023-09-23
add alternate phrasings to base search

Aprillion:
* fixing #321
* e.g. https://stampy-ui.aprillion.workers.dev/?q=simply
image

### DIFF
--- a/app/hooks/search.tsx
+++ b/app/hooks/search.tsx
@@ -1,10 +1,9 @@
 import {useState, useEffect, useRef, MutableRefObject} from 'react'
+import {Question} from '~/server-utils/stampy'
 
-export type Question = {
+export type SearchResult = {
   pageid: string
   title: string
-}
-export type SearchResult = Question & {
   score: number
   model: string
   url?: string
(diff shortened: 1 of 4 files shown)


app/hooks/search.tsx
```diff
@@ -97,7 +96,9 @@ export const baselineSearch = async (
 const normalize = (question: string) =>
   question
     .toLowerCase()
+    .replace(/\n/g, ' ')
```
mruwnik, line 99: This could cause problems with things like:
```
{
   title: "this is the end",
   alternatePhrasings: "time to say goodbye",
}
```
when searching for "end time"

Aprillion: ah, wouldn't that example be a *good* search result tho?

I know what you mean, but if alternate phrasings should be formulated as questions, accidental match seems unlikely to me TBH - and if it happens, it might not even be undesired in most cases 🤔 

also, more motivation to release production-quality API of semantic search 😅

if we wanted to make production quality keyword search, we should also prioritize match in the title above match in alternate phrasings, and a thousand other manually crafted tweaks that I don't see as the highest priority thing to spend time on for a "temporary" feature (that already lives longer than I expected TBH)

Aprillion: that said, feel free to add commits to this PR that you will be be more happy with 😉

mruwnik: nah, as long as you're aware of it :D

mruwnik (approved):

Issue #285, opened 2023-07-22
UI facelift


ccstan99:
New [Figma prototype](https://www.figma.com/file/ipifwmO1KYfqakpl3srFIL/2023-07-11-AI-Safety-ccc-ccstan99?type=design&node-id=90-2047&mode=design) for a UI facelift to dress up existing functionality. I've left some notes for the designer to update but it's mostly all there. I'd recommend tackling the low hanging fruit first: styling colors, fonts, etc then work on other aspects as time permits. A more detailed overhaul is in progress

Notes (some might need to be added as separate issues, LMK):
- Header with logo, search bar, and icon/menus is sticky. Footer should only appear below last question.
- Clicking on a tag icon refreshes page to display answers that match the tag. The behavior between clicking on a tag on an answer or from the the big tags page should be the same.
- The editors requested that tooltip with brief description appears when hovering over a tag
- If the related questions lines are shown, questions can "fade in" instead of "slide in" to minimize jarring animation. This may cause complications for nested questions or expanding related questions with their own related question.
- "Load more question..." @plexish wants it to turn on infinite loading / scrolling. I've informally sampled a few users who expect to just load a batch of another 5-10 questions, so we may need to investigate further.
- LessWrong people suggested making text window narrower for easier reading. Our current width is 1900 px ~120 characters/line. General wisdom is ~80 characters/line. I did some quick calculations and think we can (1) drop width to 1600px (2) increase left & right padding by 50px (3) increase text size a bit. This might need a bit of trial & error.

Addresses some of the concerns raised in #207

Aprillion:
Different figma from PR #327 - https://www.figma.com/file/q7DDvvkSKfYpT0yheLexht/AI-Safety-ccc-ccstan99?node-id=90%3A2047&mode=dev

ccstan99:
The Figma file linked in the opening post is the most up-to-date. We don't own or have update access to the other one used in that PR.

PR #330, opened 2023-10-01
Add unhelpful button

mruwnik:
image

One downside of this is that an article can be both liked and disliked at the same time. Though that can be a feature not a bug!


app/server-utils/stampy.ts
```diff
@@ -452,3 +452,5 @@ export const incAnswerColumn = async (column: string, pageid: PageId, subtract:
 
 export const makeColumnIncrementer = (column: string) => (pageid: PageId, subtract: boolean) =>
   incAnswerColumn(column, pageid, subtract)
+export const makeColumnDecrementer = (column: string) => (pageid: PageId, subtract: boolean) =>
+  incAnswerColumn(column, pageid, !subtract)
```
Aprillion, line 456: eeh, `!subtract`?!? 😖

TBH, I would prefer to have 2 columns in Coda to track helpful and unhelpful (`2 - 1` has a different vibe from `100 - 99`), but if it's already decided that we want only 1 column please get rid of the `subtract` parameter from all places, replacing it with a `decAnswerColumn` function (or some better way)

app/routes/questions/actions.tsx
Aprillion, line 1: 🤔 I am not a fan of the different button widths ... Figma from #285 does not show the button text, only the icons

=> @mruwnik I wonder if you could remove visible text and only use title for hover tooltips, please? or should we track in separate ~~Jira~~ GH issue?

@ccstan99 - Figma doesn't specify order for 3 buttons, is the order `👍👎✏️` going to be uncontroversial (and/or already approved from meetings) or do we want to discuss other options on Discord? (`👍✏️👎`, `✏️👍👎`, `✏️👎👍`, `👎👍✏️`, `👎✏️👍`)

Aprillion (changes requested):

app/routes/questions/actions.tsx
mruwnik: Separate issue, probably?

app/server-utils/stampy.ts
mruwnik: There are now 2 columns

Aprillion (approved):

PR #333, opened 2023-10-05
Fix manual dispatch of deploy

Aprillion:
* actions after merging PRs from forked repos don't have access to GH secrets and fail => manual deploy is needed for those
* tested in https://github.com/StampyAI/stampy-ui/actions/runs/6423175192


mruwnik (approved):
I assume this works?

PR #337, opened 2023-10-07
split test steps without duplicate build, fix lock file

Aprillion:


Aprillion:
FYI @jrhender

.github/workflows/pr-check.yml
```diff
@@ -15,7 +15,8 @@ jobs:
         run: |
           npm ci
           npm run lint
-          npm run build
```
jrhender, line 18: Is this because the `test` script is currently `npm run build && jest`, so `npm run build` is being run twice? I think this is ok as an optimization, however my concern would be that if the `test` script changes, then the CI is no longer running a build.

package-lock.json
```diff
@@ -49,7 +49,7 @@
         "wrangler": "^2.10.0"
       },
       "engines": {
-        "node": ">=16"
+        "node": ">=18"
```
jrhender, line 52: Ah thanks, perhaps I forgot to run `npm install` to update package-lock after increasing the engine version in package.json 🤔 ?

.github/workflows/pr-check.yml
Aprillion: well, if the test fails because of not running a build, we will notice 😅

package-lock.json
Aprillion: yup, fortunately no other changes were missing

PR #342, opened 2023-10-24
random redirects to random question

mruwnik:
Rob wanted the random page to work as a redirect


Aprillion (approved):
this even simplifies the code, nice

PR #343, opened 2023-10-25
POC ToC

mruwnik:
based on [this doc](https://docs.google.com/document/d/1eyHgAxiSji4FKh9tYYGvmXb24HnwtKC11IlJEwfwt_U/edit)

image

An interactive version deployed to: https://stampy-ui.ai688.workers.dev/toc

I couldn't be bothered to keep the navigation bar floating on the left, so for now it's just a normal element. Should be ok, seeing as this is just to check if it makes sense?

### DIFF
--- a/app/routes/toc.tsx
+++ b/app/routes/toc.tsx
@@ -1,8 +1,8 @@
-import {useEffect, useState} from 'react'
+import {useEffect, useState, MouseEvent} from 'react'
 import {useOutletContext} from '@remix-run/react'
 import {Header, Footer} from '~/components/layouts'
 import useQuestionStateInUrl from '~/hooks/useQuestionStateInUrl'
-import {fetchQuestion, Question} from '~/routes/questions/$question'
+import {LINK_WITHOUT_DETAILS_CLS, fetchQuestion, Question} from '~/routes/questions/$question'
 import {QuestionState} from '~/server-utils/stampy'
 import type {Question as QuestionType, PageId} from '~/server-utils/stampy'
 import type {Context} from '~/root'
@@ -42,7 +42,7 @@ const ChildList = ({items, selectItem}: {items: Item[]; selectItem: (i: Item) =>
       {items.map((child, i) => {
         if (child?.items && child.items.length > 0) {
           return (
-            <details key={i} id={normalizeName(child?.content)} open>
+            <details key={i} id={normalizeName(child?.content)}>
               <summary>
                 <MaybeURL item={child} selectItem={selectItem} />
               </summary>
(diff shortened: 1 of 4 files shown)


app/routes/toc.tsx
```diff
@@ -0,0 +1,112 @@
...
+    const fetcher = async () => {
+      const response = await fetch('/toc.json')
+      setToc(await response.json())
+    }
+    fetcher()
+  }, [])
+
+  return (
+    <>
+      <style>{`
```
mruwnik, line 101: this is fun :D

app/routes/toc.tsx
```diff
@@ -0,0 +1,112 @@
...
+        body { max-width: 1400px; }
+            `}</style>
```
Aprillion, line 103:
```suggestion
      <style>
        body { max-width: 1400px; }
      </style>
```

app/routes/toc.tsx
```diff
@@ -0,0 +1,112 @@
...
+  return (
+    <article className="toc" style={{padding: '10px'}}>
+      {toc.map((main: Item, i: number) => (
+        <Section key={i} item={main} />
+      ))}
+    </article>
+  )
+}
+
+export default function App() {
```
Aprillion, line 88:
```suggestion
export default function Toc() {
```

public/toc.json
Aprillion, line 1: were any plans discussed where to host the JSON file long-term? or would it change rarely enough to keep it in git?

Aprillion (approved):

public/toc.json
mruwnik: so far it was mainly to just see if this even makes sense - my assumption is that if it does, then it's probably best for it to be a Coda column (e.g. `parent`) and have the UI backend construct the map of questions. Or something like that

mruwnik:
changed to:

image

ccstan99 (commented):
This layout looks really promising! A few suggestions that should hopefully be easy fixes:
- Text on navigation menu should all be same size. There's no need for headers to be larger since the organization is easily discernible now
- I would start all menu items collapsed by default for cleaner, less overwhelming visual display
- Instead of starting with a blank page, pick an answer to display in the main panel by default, maybe "What is AI safety?" Eventually, this could be some kind of welcome / overview text.

app/routes/toc.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  return (
+    <>
+      <style>{`
+        body { max-width: 1400px; }
+            `}</style>
+      <Header />
+      <main className="toc-container" onClick={handleSpecialLinks}>
+        <ToC toc={toc} selectItem={selectItem} />
+        <div style={{flex: '5 3 auto'}}>
```
Aprillion, line 153: I would use either `.toc-container {...}` in global styles or use inline styles, but let's not combine both approaches in the same compoment, please

Aprillion (approved):

app/routes/toc.tsx
mruwnik: agreed, but can't be bothered to now :P

PR #346, opened 2023-11-04
make all links from the answers external #340

### LINKED ISSUE #340: AISafety.com integration: Clicking internal links switches out of embed mode

Aprillion:
* fix #340 by making all links from the answers external
* remove the external link icon if we don't need to distinguish between 2 types of links
* https://stampy-ui.aprillion.workers.dev/

### DIFF
--- a/app/root.css
+++ b/app/root.css
@@ -818,19 +818,6 @@ a.see-more.visible:after {
   content: 'See less';
 }
 
-a[target='_blank']:not(.icon-link, .transparent-link):after {
-  content: ' ';
-  display: inline-block;
-  vertical-align: top;
-  width: 12px;
-  height: 12px;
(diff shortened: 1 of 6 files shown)


mruwnik:
is this how it should always work? So search and related questions will add new cards, but links in the question will reload the site? That might be a bit confusing...

Aprillion:
> is this how it should always work? So search and related questions will add new cards, but links in the question will reload the site? That might be a bit confusing...

what reload? it should always open on new tab, can you give me repro steps when it reloads?

mruwnik:
`s/reload/new tab/g`. But the functional result is the same. Can be good, I suppose, but would be unexpected for me

mruwnik (approved):

Issue #329, opened 2023-10-01
Tags link behavior


ccstan99:
Clicking on a tag from the main page should have same behavior as clicking on a tag from the tags page. Show a page with all questions matching tags on page instead of popping up a dialogue box.

PR #347, opened 2023-11-09
Change tags buttons to links

mruwnik:
https://github.com/StampyAI/stampy-ui/issues/329

### DIFF
--- a/app/routes/tags/$tag.tsx
+++ b/app/routes/tags/$tag.tsx
@@ -1,7 +1,7 @@
-import {useState, ReactNode} from 'react'
+import {useState, useEffect, ReactNode} from 'react'
 import {LoaderFunction} from '@remix-run/cloudflare'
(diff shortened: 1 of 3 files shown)


app/routes/tags/$tag.tsx
```diff
@@ -41,6 +41,36 @@ export async function fetchTag(tagName: string): Promise<TagType | never[]> {
...
+}: {
+  name: string
+  questions?: RelatedQuestions
+  showCount?: boolean
+}) {
+  const [questions, setQuestions] = useState(tqs)
+  const pageIds = questions?.map((q) => q.pageid).join(QuestionState.COLLAPSED)
+
+  useEffect(() => {
+    const fetcher = async () => {
```
mruwnik, line 57: the obvious downside of this is that there will be loads of extra queries flying around

Aprillion: if we have them cached in CF, it should be fine 🤞

Aprillion (approved):

PR #386, opened 2024-02-10
pin @node/types to 20.8.3

jrhender:
I noticed that stampy.ts tests are failing (e.g. https://github.com/StampyAI/stampy-ui/actions/runs/7854687158/job/21435971107) with the following error:
```
    app/server-utils/stampy.ts:196:45 - error TS2576: Property 'json' does not exist on type 'Response'. Did you mean to access the static member 'Response.json' instead?

    196     json = await (await fetch(url, params)).json()
```
I think it is due to the types conflict described [here](https://github.com/cloudflare/workerd/issues/1298).

This PR removes `@types/node` to remove the conflict. This doesn't seem to break the app or any tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,6 @@
         "@types/lodash": "^4.14.202",
         "@types/markdown-it": "^13.0.7",
         "@types/markdown-it-footnote": "^3.0.3",
-        "@types/node": "20.8.3",
         "@types/react": "^18.2.53",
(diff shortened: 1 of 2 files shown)


Aprillion (approved):
IMHO it was probably due to stale node_modules - happened to me a few times and deleting them and reinstalling helped for me, so I guess any change in package.json "fixed" it for you :D

PR #393, opened 2024-02-13
[Ticket ID or name] Change button background to red

melissasamworth:
Any extra comments such as why


mruwnik (approved):
this is a comment. Fix something

app/components/Button/button.css
```diff
@@ -12,7 +12,8 @@
 }
 
 .primary {
-  background: var(--colors-teal-500, #1d9089);
+  /* background: var(--colors-teal-500, #1d9089); */
+  background: red;
```
mruwnik, line 16: why red, rather than orange?

melissasamworth: Not sure. Maybe Illl try orange

PR #399, opened 2024-02-14
style: article main content&buttons styling

buddy-web3:


Aprillion (approved):

app/components/Article/article.css
```diff
@@ -103,3 +104,7 @@ article .footer-comtainer {
 article .footer-comtainer > * {
   margin: var(--spacing-8, 8px);
 }
+
+h1.teal {
```
mruwnik, line 108: just add the `teal-500` class to the h1

app/components/Article/article.css
```diff
@@ -45,6 +45,7 @@ article .tag:hover {
 
 article .contents a {
   color: var(--colors-teal-500);
+  font-weight: bolder;
```
melissasamworth, line 48: No font properties should ever be specified in the CSS because we have type classes in newroot.css, always just apply one of those.

Can we only use font-weight: 400 or font-weight: 700?

app/components/Button/button.css
```diff
@@ -97,6 +97,7 @@
   border-radius: 5px;
   display: none;
   z-index: 1;
+  font-size: 16px;
```
melissasamworth, line 100: CSS looks fine, but can we make tooltip its own component?

Issue #379, opened 2024-02-09
categories page


melissasamworth:
**Update (12 Feb)**
Almost done, just a few small styling issues (see comment below)

### Acceptance Criteria
- [ ] styling should match figma
- [x] navigating to `/tags` should navigate to default tag. This is important for backwards compatibility because https://aisafety.info/tags works on current site.
- [x] URLs should be `tag/{tag id}/{tag name}`
  - Navigating to `tag/{tag id}/anything` should navigate to correct tag using id, but will try to append name where possible

jrhender:
@melissasamworth @mruwnik for the article page, issue #382 asks that the links be "aisafety.com/id/current-title-of-article". Would it make sense to also have the category URLs have the name and the id? In the current implementation, the category URLs are `/tags/{tag name}`. Would it instead make sense to do `/tags/{tag id}/{tag name}`, similar to the articles?

jrhender:
This tasks probably isn't quite done from a styling standpoint as the sidebar doesn't quite match figma:
- It isn't aligned under the "AISafety.info" logo
- The green highlight on the selected category isn't correct

![Image](https://github.com/StampyAI/stampy-ui/assets/4407464/9827a40c-c8d2-4760-b758-758d0fc7084a)

melissasamworth:
Hey @jrhender quick process note: do you think it's best to talk about individual tickets here on Github? I think it probably does, and if so I'll make sure to check this daily (sorry for the lag this time).

The purpose of the IDs with articles is to make it backwards compatible, particularly for links people have shared that are floating out there. I'm not sure how many tag links people have shared. Still, backwards compatibility would be nice. Does your solution solve that problem?

jrhender:
> The search input current says "Search articles" but it should say "Filter by keyword"

PR here: https://github.com/StampyAI/stampy-ui/pull/391

jrhender:
> Lastly, not sure if it should belong to this issue but the widths of the layout is off. Ideally they should be the following viewport widths

@melissasamworth @buddy-web3 

Do you think the change to fix this would be in `newRoot.css`? https://github.com/StampyAI/stampy-ui/blob/04b76d21a98c0470814a78260a896f8f1e359f59/app/newRoot.css#L408

melissasamworth:
Storybook version pushed
Still needs to be integrated with article page (Nemo)

jrhender:
Hey @melissasamworth , could you elaborate on this point please:
> Storybook version pushed

[...]

PR #402, opened 2024-02-14
feat(menu.module.css): padding to category to fix active highlighting

jrhender:
Relates to #379 . Adds padding so that active category is closer to figma

image

### DIFF
--- a/app/components/CategoriesNav/menu.module.css
+++ b/app/components/CategoriesNav/menu.module.css
@@ -15,6 +15,7 @@
   align-items: flex-start;
   margin: 5px;
   position: relative;
+  padding: 0px 10px 0px 24px;
 }
 .categoryAutoLayoutHorizontal.active {
   background-color: rgba(237.0000010728836, 250.00000029802322, 249.00000035762787, 1);


app/components/CategoriesNav/menu.module.css
```diff
@@ -15,6 +15,7 @@
   align-items: flex-start;
   margin: 5px;
   position: relative;
+  padding: 0px 10px 0px 24px;
```
melissasamworth, line 18: We're trying to use variables when we define numbers. Partially for conversion to mobile, partially so we can avoid numbers that aren't in the  design system like 10px. If you want you can change these but I'll still approve :).

melissasamworth (approved):

app/components/CategoriesNav/menu.module.css
jrhender: Ok, thanks. I'll adjust this PR to have variables and numbers that are in the design system.
If you want to merge to get the change on `stampy-redesign` though, go ahead and I can make the change in another PR.

jrhender:
Closing as categories page is in a good state AFAIK

PR #408, opened 2024-02-15
Deploy test env

mruwnik:
I'm not sure how to best deploy a test version of the new site, so please criticize this

### DIFF
--- a/.github/workflows/deploy-dev.yml
+++ b/.github/workflows/deploy-dev.yml
@@ -1,0 +1,32 @@
+name: Deploy to dev Cloudflare Worker
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - stampy-redesign
+    types:
+      - closed
+env:
+  CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+jobs:
+  deploy:
+    if: github.event.pull_request.merged == true || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone stampy-ui repository code
+        uses: actions/checkout@v3
+      - name: Deploy to Cloudflare
+        run: |
+          cat wrangler.toml.template \
+            | sed s/stampy-ui/redesign/ \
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-dev.yml
```diff
@@ -0,0 +1,32 @@
+name: Deploy to dev Cloudflare Worker
```
mruwnik, line 1: I considered parametrizing the current build script, but thought it better to have separate ones

.github/workflows/deploy-dev.yml
```diff
@@ -0,0 +1,32 @@
...
+  deploy:
+    if: github.event.pull_request.merged == true || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone stampy-ui repository code
+        uses: actions/checkout@v3
+      - name: Deploy to Cloudflare
+        run: |
+          cat wrangler.toml.template \
+            | sed s/stampy-ui/redesign/ \
```
mruwnik, line 21: will this suffice for it to not overwrite the current page?

Aprillion: no idea 🤷

.github/workflows/deploy-dev.yml
Aprillion: that sounds better to me too 😅

.github/workflows/deploy-dev.yml
Aprillion: but I created 2 new GH secrets if you want to replace next 2 lines to be on the safe side 😅 

then you can try that https://stampy-ui.aprillion.workers.dev/ is the version from my manual deploy last week vs https://redesign.aprillion.workers.dev/ would be created from this action (feel free to add some version indicator, e.g. printing git commit date in console or something, if it would be hard to spot a difference)

```
CLOUDFLARE_ACCT_ID_APRILLION_WORKERS_DEV
STAMPY_KV_ID_APRILLION_WORKERS_DEV
```

let me know if there are more steps to enable something somewhere to allow deploy from GH action in my CF account that I forgot about

jrhender (approved):
LGTM

PR #412, opened 2024-02-16
Stampy redesign 368

buddy-web3:


app/components/Button/button.css
```diff
@@ -107,3 +107,23 @@
 .tooltip:hover::after {
   display: block;
 }
+
+.action-feedback-text {
+  display: none;
+  position: absolute;
```
mruwnik, line 113: can you do this with relative values? This will not work on smaller screens

app/components/Article/index.tsx
```diff
@@ -35,8 +38,25 @@ const ArticleFooter = (question: Question) => {
...
+            onSuccess={() => setShowFeedback(true)}
+          />
+          <Action
+            pageid={question.pageid}
+            showText={true}
+            actionType={ActionType.UNHELPFUL}
+            onSuccess={() => setShowFeedbackForm(true)}
+          />
+          <span className={['action-feedback-text', showFeedback ? 'show' : ''].join(' ')}>
+            Thanks for your feedback!
```
mruwnik, line 54: @melissasamworth  what will cause this to be hidden again? Should it disappear e.g. after 10s? Otherwise it will hang around and be annoying

app/components/Article/FeedbackForm/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+          return {...feedback, selected: !feedback.selected}
+        }
+        return {...feedback, selected: false}
+      })
+    )
+    setEnabledSubmit(true)
+  }
+
+  return (
+    <div className={className}>
```
mruwnik, line 48: why not just make this into a html form? A radio input + button should handle most of this

app/components/Article/FeedbackForm/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+          >
+            <div className={'small'}>{option.text}</div>
+          </div>
+        ))}
+        <textarea
+          name={'feedback-text'}
+          className={'feedback-text'}
+          placeholder={'Leave a comment (optional)'}
+        ></textarea>
+        <div className={['submit-feedback', enabledSubmit ? 'enabled' : ''].join(' ')}>
```
mruwnik, line 65: use the `Button` component

app/components/Article/FeedbackForm/feedbackForm.css
```diff
@@ -0,0 +1,88 @@
+.feedback-header {
+  color: var(--colors-black);
+}
+.feedback-container {
+  border-radius: 6px;
+  background-color: #fff;
+  box-shadow: 0px 16px 40px rgba(175, 183, 194, 0.2);
+  border: 1px solid #dfe3e9;
```
melissasamworth, line 8: could you reference the var?

app/components/Article/FeedbackForm/feedbackForm.css
```diff
@@ -0,0 +1,88 @@
...
+  box-sizing: border-box;
+  width: 384px;
+  height: 542px;
+  padding: var(--spacing-32);
+}
+.select-option {
+  border-radius: 6px;
+  background-color: #fff;
+  box-shadow: 0px 16px 40px rgba(175, 183, 194, 0.2);
+  border: 1px solid #dfe3e9;
```
melissasamworth, line 18: here too

As a general rule, too, if we're repeating code a lot we should just create a one-off class for newroot.css like

.border {
 border: 1px solid #dfe3e9;
}

app/components/Article/FeedbackForm/feedbackForm.css
```diff
@@ -0,0 +1,88 @@
...
+.select-option.selected {
+  border: 1px solid var(--colors-teal-500);
+}
+
+.feedback-form {
+  width: 100%;
+  position: relative;
+  height: 542px;
+  text-align: left;
+  font-size: 16px;
```
melissasamworth, line 39: The idea was to make it so you never have to write a line of code related to anything type. Just look at what class I use in Figma and find that same class in newroot.css

image

app/components/Article/FeedbackForm/feedbackForm.css
```diff
@@ -0,0 +1,88 @@
...
+  color: var(--colors-cool-grey-600);
+  font-family: Poppins;
```
melissasamworth, line 41: same as comment above :)

app/components/Article/FeedbackForm/feedbackForm.css
```diff
@@ -0,0 +1,88 @@
...
+  max-width: calc(100% - var(--spacing-24));
+  height: 128px;
+  border-radius: 6px;
+  box-shadow: 0px 16px 40px rgba(175, 183, 194, 0.2);
+  border: 1px solid var(--colors-cool-grey-200);
+  padding: var(--spacing-12);
+}
+.small {
+  position: relative;
+  letter-spacing: -0.01em;
```
melissasamworth, line 56: same again heh

app/components/Article/FeedbackForm/feedbackForm.css
```diff
@@ -0,0 +1,88 @@
...
+  line-height: 170%;
```
melissasamworth, line 57: and here

app/components/Article/FeedbackForm/feedbackForm.css
```diff
@@ -0,0 +1,88 @@
...
+}
+
+.small3 {
+  position: relative;
+  letter-spacing: -0.01em;
+  line-height: 170%;
```
melissasamworth, line 63: ++

app/components/Article/FeedbackForm/feedbackForm.css
```diff
@@ -0,0 +1,88 @@
...
+  font-weight: 600;
```
melissasamworth, line 64: ++

app/components/Article/FeedbackForm/feedbackForm.css
```diff
@@ -0,0 +1,88 @@
...
+  border-radius: 6px;
+  background-color: var(--colors-teal-500);
+  box-shadow: 0px 16px 40px rgba(32, 44, 89, 0.05);
+  width: 320px;
+  height: 48px;
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  justify-content: center;
+  padding: 7px 17px;
```
melissasamworth, line 79: Always use a variable. If the padding in the file was 7px I meant for it to be 8px sorry! 17px would be 16px.

app/components/Article/index.tsx
melissasamworth: Yeah that works, disappears after 10s

melissasamworth (changes requested):

app/components/Article/FeedbackForm/index.tsx
buddy-web3: all other comments have been addressed. this one is slightly more css-nasty.

app/components/Button/index.tsx
```diff
@@ -8,8 +8,9 @@ type ButtonProps = {
   className?: string
   tooltip?: string
   icon?: ReactNode
+  disabled?: boolean
 }
-const Button = ({children, action, tooltip, icon, className}: ButtonProps) => {
+const Button = ({children, action, tooltip, icon, className, disabled = false}: ButtonProps) => {
```
mruwnik, line 13: how would this work for link buttons? You can add a `onClick={(e) => e.preventDefault()}` or something, to disable it, but it might require a bit extra styling

app/components/Button/button.css
```diff
@@ -107,3 +107,34 @@
 .tooltip:hover::after {
   display: block;
 }
+
```
mruwnik, line 110: could you move the feedback specific items to feedbackForm.css?

app/components/Article/FeedbackForm/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  const handleSubmit = () => {}
+
+  return (
+    <div className={className} onBlur={handleBlur} onFocus={onFocus}>
+      <div className={'feedback-container bordered'}>
+        <span className={'feedback-header'}>What was the problem?</span>
+        {hasOptions
+          ? feedbackOptions.map((option, index) => (
+              <div
```
mruwnik, line 93: aren't these also buttons? So something like `<Button className={option.selected ? 'secondary-alt bordered' : 'secondary'} action={() => selectFeedback(index)}>{option.text}</Button>`

app/components/Article/FeedbackForm/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+    },
+    {
+      text: 'Typos',
+      selected: false,
+      option: 'typos',
+    },
+  ])
+  const [enabledSubmit, setEnabledSubmit] = React.useState(false)
+  const selectFeedback = (option: number) => {
+    setFeedbackOptions(
```
mruwnik, line 55: why not just store the selected item? So:
```
const [selected, setSelected] = useState<string>()

const selectFeedback = (option: string) => {
    setSelected(option)
}
(...)
<div className={option.option === selected ? 'selected' : ''}
    onClick={() => selectFeedback(option.option)}>
  {option.text}
</div>
```

app/components/Article/index.tsx
```diff
@@ -8,17 +8,33 @@ import {Action, ActionType} from '~/routes/questions.actions'
...
+  const [showFeedbackForm, setShowFeedbackForm] = useState(false)
+  const [isFormFocused, setIsFormFocused] = useState(false)
   const date =
     question.updatedAt &&
     new Date(question.updatedAt).toLocaleDateString('en-GB', {
       year: 'numeric',
       month: 'short',
     })
 
+  React.useEffect(() => {
```
mruwnik, line 26: why not move all this stuff into the feedback form component? It's only hidden, so the HTML elements will still be around

app/components/Article/index.tsx
mruwnik: this also should probably go into the feedback form

app/components/Article/FeedbackForm/feedbackForm.css
```diff
@@ -0,0 +1,44 @@
+.feedback-header {
+  color: var(--colors-black);
```
mruwnik, line 2: you can just use the `black` class for this

app/components/Button/index.tsx
buddy-web3: is it necessary for anything else besides the feedback form?

app/components/Article/index.tsx
buddy-web3: for storybook functionality

app/components/Button/index.tsx
mruwnik: not now, but it would be good to have it done for consistency

app/components/Article/index.tsx
mruwnik: that's the wrong way round. Storybook, tests etc. shouldn't change how the actual code is structured. If something is hard to do without changes, then it's a hint that something is wrong, but the code should dictate storybook, not the other way round

mruwnik: You can add extra wrappers with decorators - check the ArticlesKeepGoing story

buddy-web3: I've been looking into this but I don't believe it's a feasible option. The article makes it display  when submitted, moving part of the logic to the Feedback component would just duplicate logic and code.

app/components/Button/button.css
```diff
@@ -104,6 +104,13 @@
   z-index: 1;
 }
 
-.tooltip:hover::after {
-  display: block;
+.button.full-width {
+  width: 100%;
+  height: 48px;
```
melissasamworth, line 109: I would say use the button component and apply the relevant class

app/components/Button/button.css
```diff
@@ -104,6 +104,13 @@
...
+  border: 0;
+}
+.button:disabled,
```
melissasamworth, line 112: these kinds of things should be added to the button component

melissasamworth (changes requested):

melissasamworth (approved):

PR #414, opened 2024-02-16
lots of CSS changes

melissasamworth:


app/components/Article/article.css
```diff
@@ -45,15 +52,19 @@ article .tag:hover {
...
-  font-weight: bolder;
+  font-weight: 700;
 }
 
 article .contents a:visited {
   color: var(--colors-teal-800);
 }
 
+article .contents p {
+  padding-bottom: 32px;
```
mruwnik, line 63: maybe a var here...

app/components/Article/index.tsx
```diff
@@ -101,9 +101,9 @@ export const Article = ({question, glossary}: ArticleProps) => {
   const {title, text, pageid} = question
 
   return (
-    <article className="article-container">
-      <h1 className="teal">{title}</h1>
-      <ArticleMeta {...question} />
+    <article>
+      <h1 className="teal padding-bottom-24">{title}</h1>
+      <ArticleMeta question={question} className="padding-bottom-56"/>
```
mruwnik, line 106: you could just add this directly to the `ArticleMeta` - it's only used here. This is also fine, though

mruwnik (approved):

app/components/Article/index.tsx
melissasamworth: This took me a lot of hard work that is not being appreciated. Ha but no I did realize that but did it anyway for practice.

PR #444, opened 2024-02-22
landing page mobile

mruwnik:


app/components/ContentBox/box.css
```diff
@@ -52,3 +52,15 @@
   padding: 0;
   padding-right: var(--spacing-8);
 }
+
+@media (max-width: 600px) {
+  .main-container-box-table {
+    flex-direction: column;
+    height: 550px;
+    margin-bottom: var(--spacing-40);
```
melissasamworth, line 60: Looks good but remember that --spacing-40 turns into I think 24px on mobile, so hope you were trying to make this appear 24px!
image

app/components/Footer/footer.css
```diff
@@ -10,3 +10,11 @@
   align-items: baseline;
   margin: var(--spacing-104) 13.33vw var(--spacing-80) 13.33vw;
 }
+
+@media (max-width: 600px) {
+  .footer-contents {
+    flex-direction: column;
+    gap: var(--spacing-32);
```
melissasamworth, line 17: I think using spacing-80 var, making is 40px for mobile. Sorry for my not super organized system, will try to make it better soon.

app/components/Grid/grid.css
```diff
@@ -1,9 +1,16 @@
...
+  flex-flow: wrap;
+}
+
 .grid-item {
   cursor: pointer;
   text-align: left;
   padding: var(--spacing-32) var(--spacing-32);
-  min-width: var(--spacing-288);
-  min-height: var(--spacing-288);
+  width: var(--spacing-288);
```
melissasamworth, line 12: I think these should be based on viewport width. I generally won't use variables to define content that is not viewport width (e.g. popovers).

app/newRoot.css
```diff
@@ -304,6 +304,9 @@ ul {
   flex-grow: 2;
 }
 
+.centered {
```
melissasamworth, line 307: thanks for using more one-off class names

app/newRoot.css
```diff
@@ -348,3 +351,53 @@ ul {
...
+  display: inherit;
+}
+
+.mobile-only {
+  display: none;
+}
+
+@media (max-width: 600px) {
+  .page-body {
+    padding: 0px 6.333vw;
```
melissasamworth, line 365: cool yeah, and let's make sure we use `.page-body` everywhere that uses standard margins, including the footer actually

melissasamworth (changes requested):

app/newRoot.css
mruwnik: I actually wanted to delete this before pushing :/ I only had it for testing...

app/components/Grid/grid.css
mruwnik: won't that get hairy with very small screen? Also, do you have specific values? :D

app/components/ContentBox/box.css
```diff
@@ -57,3 +57,15 @@
   padding: 0;
   padding-right: var(--spacing-8);
 }
+
+@media (max-width: 600px) {
```
melissasamworth, line 61: We're using 640

app/newRoot.css
```diff
@@ -348,3 +351,49 @@ ul {
...
+
+.desktop-only {
+  display: inherit;
+}
+
+.mobile-only {
+  display: none;
+}
+
+@media (max-width: 600px) {
```
melissasamworth, line 363: 640px

melissasamworth (approved):

PR #446, opened 2024-02-23
fixes

mruwnik:


app/components/Article/Contents.tsx
```diff
@@ -123,9 +123,9 @@ const insertGlossary = (pageid: string, glossary: Glossary) => {
       addPopup(
         e as HTMLSpanElement,
         `glossary-${entry.term}`,
-        `<div class="glossary-popup flex-container black col-6 small">
+        `<div class="glossary-popup flex-container black small">
               <div class="contents col-8">
-                   <h3>${entry.term}</h3>
+                   <div class="small-bold">${entry.term}</div>
```
mruwnik, line 128: @melissasamworth you had `p` on the bug report, but that has auto padding set from the underlying article, so I left divs

app/components/Article/article.css
```diff
@@ -96,6 +97,7 @@ article .defintion {
 }
 article .link-popup .glossary-popup > img {
   width: 100%;
+  height: 303px;
```
mruwnik, line 100: the image was covering the bottom border, so I chopped off a pixel here

app/components/Article/Contents.tsx
melissasamworth: That's fine. In the future perhaps we can create a `padding: 0` class to add or something

melissasamworth (approved):

melissasamworth (approved):

PR #447, opened 2024-02-23
mobile global nav

buddy-web3:


app/components/Button/index.tsx
```diff
@@ -8,20 +8,24 @@ type ButtonProps = {
...
   icon?: ReactNode
+  leftIcon?: ReactNode
 }
-const Button = ({children, action, tooltip, icon, className}: ButtonProps) => {
+const Button = ({children, action, tooltip, icon, leftIcon, className}: ButtonProps) => {
   const classes = ['button', className, tooltip && 'tooltip'].filter((i) => i).join(' ')
   if (typeof action === 'string') {
     return (
       <Link to={action} className={classes} data-tooltip={tooltip}>
+        {leftIcon && leftIcon}
```
mruwnik, line 18: I don't like how this is getting bloated. Why not just pass the icon in as a child, rather than having special parameters? So e.g.:
```
<Button>
   <SomeIcon />
   bla bla
</Button>
```
rather than
`<Button leftIcon={<SomeIcon/>}>bla bla</Button>`?

app/components/Nav/Mobile/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+          <nav className="articles-header">
+            <CarrotLarge className={'back-icon'} onClick={() => setShowArticles(false)} />
+            <p className={'composite-elements'}>
+              <OpenBookIcon className={'icon-margin'} />
+              Articles
+            </p>
+            <XLarge className={'menu-button'} onClick={toggleMenu} />
+          </nav>
+
+          <div className={'articles-mobile'}>
```
mruwnik, line 71: this should reuse the `ArticlesDropdown` component. A bit of extra css there should handle transforming it from a grid into a list

app/components/Nav/Mobile/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const toggleMenu = () => {
+    setShowMenu(!showMenu)
+    setShowArticles(false)
+  }
+  console.log('toc', toc)
+  return (
+    <header className="top-header">
+      {!showArticles && (
+        <>
+          <nav className="top-nav">
```
mruwnik, line 28: how much of this could be handled with css in the normal `Nav` component?

buddy-web3: I'm not sure I follow, can you please explain what you mean?

app/components/Nav/Mobile/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+              .map((item: TOCItem) => (
+                <Link
+                  key={`${item.pageid}-${item.title}`}
+                  to={questionUrl(item)}
+                  className="articles-entry"
+                >
+                  {item.title}
+                </Link>
+              ))}
+            <div className="default-bold margin-bottom top-margin">Browse by category</div>
```
melissasamworth, line 96: what are the classes margin-bottom and top-margin? For bottom padding we want to use the class "padding-bottom-[number]". Maybe I will create padding-top classes and even margin classes too...

app/components/Nav/Mobile/navMobile.css
```diff
@@ -0,0 +1,59 @@
+@media only screen and (max-width: 640px) {
+  .top-header {
+    padding: 0;
+  }
+  .top-nav {
+    justify-content: space-between;
+    padding: var(--spacing-20) var(--spacing-24);
```
melissasamworth, line 7: The nav was supposed to be based on height (using 80px) rather than padding. If you want to use padding can you just round to the nearest one, so either 16px or 24px? And remove the --spacing-20 var

app/components/Nav/Mobile/navMobile.css
```diff
@@ -0,0 +1,59 @@
...
+  }
+  .space-between {
+    justify-content: space-between;
+  }
+  .flex-start {
+    justify-content: flex-start;
+  }
+  .mobile-menu > .button {
+    border-radius: 0;
+    font-size: 16px;
```
melissasamworth, line 27: never write any CSS for fonts! Assign whichever is the right class based on the Figma, or just h1 or h2 tag.

app/newRoot.css
```diff
@@ -32,6 +32,7 @@
   --spacing-8: 8px;
   --spacing-12: 12px;
   --spacing-16: 16px;
+  --spacing-20: 20px;
```
melissasamworth, line 35: remove this, sorry!

melissasamworth (changes requested):

app/components/Nav/Mobile/index.tsx
mruwnik: I only just noticed that I didn't send this comment...

mruwnik: preferably. But this certainly still needs to be connected to the actual page

mruwnik (approved):

Issue #420, opened 2024-02-19
mobile - make sure every single tag that contains type has the correct class


melissasamworth:
**Acceptance criteria:**

- Every single H tag should have no classes assigned other than color classes or padding classes
- Each tag should match the tag specified in the design
- All p tags should be assigned a class (the class specified in the design)
![Image](https://github.com/StampyAI/stampy-ui/assets/825910/572b6c32-0d67-46cf-86ba-2b7c0b2ef4cd)
- The class specified in the design matches to one of these classes in root.css this should be easy
![Image](https://github.com/StampyAI/stampy-ui/assets/825910/94691373-92c9-4501-9d8a-2b958c7a29be)
- Make sure no extra classes are assigned to p tags other than color classes or padding classes
- Make sure no typography styling is **ever defined anywhere else in the CSS**

This is all important for conversion to mobile, as well as just consistency and making our process faster!

jrhender:
From a quick search through the code for all `<h`, I think all of the `h` tags have no classes assigned other than colour classes or padding classes now. @melissasamworth , is it okay if an `h` tag doesn't have colour and/or padding classes?

jrhender:
@melissasamworth I think the only `p` tag that doesn't have the figma classes is this tag in `questions.add`: https://github.com/StampyAI/stampy-ui/blob/stampy-redesign/app/routes/questions.add.tsx#L86
Is this route to add a question in the new design yet?

melissasamworth:
@jrhender yes an H without classes is fine

In the case that the type is the default paragraph type it won't have a class so this is probably okay. I think we're good if that was the last thing you found.

PR #448, opened 2024-02-23
fix(_index): remove large-bold class from Advanced content <h> tag

jrhender:
According to issue #420 , tags should only have colour and padding classes


app/routes/_index.tsx
```diff
@@ -45,7 +45,7 @@ export default function App() {
         <WidgetStampy />
 
         <div className="padding-bottom-80" />
-        <h3 className="grey large-bold padding-bottom-32">Advanced content</h3>
+        <h3 className="grey padding-bottom-32">Advanced content</h3>
```
melissasamworth, line 48: Heh no so, we want it to be a p tag, I'm not using h3s.

`<p className="grey large-bold padding-bottom-32">Advanced content</p>`

melissasamworth (changes requested):

app/routes/_index.tsx
jrhender: @melissasamworth  Thanks for the note. I changed the tag to a `p`.

Should it still have the padding-bottom-32 class? How can one get the desired padding from the figma?

PR #450, opened 2024-02-26
mobile Keep Going

buddy-web3:


app/components/Article/KeepGoing/keepGoing.module.css
```diff
@@ -4,3 +4,14 @@
   padding: var(--spacing-32);
   margin-bottom: var(--spacing-56);
 }
+
+@media (max-width: 640px) {
+  .flex_dynamic {
```
jrhender, line 9: Other two word class names seem to be separated with a hyphen; probably best to stay consistent here unless there is a special reason.
```suggestion
  .flex-dynamic {
```

buddy-web3: @jrhender  In react it's imported as styles.module, hyphen is not a valid character in a JS variable

app/components/Article/KeepGoing/keepGoing.module.css
```diff
@@ -4,3 +4,14 @@
   padding: var(--spacing-32);
   margin-bottom: var(--spacing-56);
 }
+
+@media (max-width: 640px) {
```
jrhender, line 8: I noticed that this `640px` value is being used in multiple media queries? (For example, also in this PR https://github.com/StampyAI/stampy-ui/pull/451). Would it make sense to define a variable that there is only a single spot to change and make it easier to stay consistent between the different media queries?
(I got this syntax from Claude, so apologies if it is a bit off 😆 )
```suggestion
@media (max-width: var(--mobile-breakpoint)) {
```

jrhender, line 8: Why the use of `@media only screen` in this other PR: https://github.com/StampyAI/stampy-ui/pull/451, but not in this one?

buddy-web3: Honestly no idea.. my IDE (intellij idea) decided to mark it as an error and block the commit unless I put `only screen`

mruwnik: I tried that, but it didn't want to work. There might be a clever way of doing it

jrhender: Ah okay, fair enough. Unfortunately, I don't know what this clever way is then 😅

melissasamworth: :(

app/components/Article/KeepGoing/keepGoing.module.css
jrhender: Ok, thanks for the explanation!

PR #451, opened 2024-02-26
article mobile bits

buddy-web3:


app/components/Article/article.css
```diff
@@ -130,3 +130,12 @@ article .footer-comtainer > * {
...
 }
+@media only screen and (max-width: 640px) {
+  article .footer-comtainer {
+    flex-direction: row;
+    flex-wrap: wrap;
+  }
+  article .footer-comtainer > div:nth-child(-n + 2) {
+    flex: 1 1;
+  }
+}
```
jrhender, line 141: A couple small typos
```suggestion
@media only screen and (max-width: 640px) {
  article .footer-container {
    flex-direction: row;
    flex-wrap: wrap;
  }
  article .footer-container > div:nth-child(-n + 2) {
    flex: 1 1;
  }
}
```

app/components/Article/article.css
```diff
@@ -130,3 +130,12 @@ article .footer-comtainer > * {
 h1.teal {
   color: var(--colors-teal-500);
 }
+@media only screen and (max-width: 640px) {
+  article .footer-comtainer {
+    flex-direction: row;
+    flex-wrap: wrap;
+  }
+  article .footer-comtainer > div:nth-child(-n + 2) {
```
jrhender, line 138: Is it a good practice to use `div:nth-child` as used? Perhaps an additional class would be clearer when reading the code and less likely to break unexpectedly? I don't have a strong feeling here, just thought I would ask.
```suggestion
  article .footer-comtainer .footer-section {
```

My CSS knowledge is limited, so my apologies if the rationale for the current approach is obvious 😄

app/components/Article/article.css
buddy-web3: @jrhender I noticed the typo as well.. but it's everywhere, not just in my additions.

app/components/Article/article.css
buddy-web3: Both options are fine by me. @mruwnik do you prefer nth-child or a new class?

app/components/Article/article.css
jrhender: Ah I see. Better to keep as is then and fix in another PR I suppose.

melissasamworth:
All looks good to me, let's wait on a final decision re "div:nth-child" before merging

melissasamworth (approved):

PR #458, opened 2024-02-29
mobile header on all pages.

buddy-web3:


app/components/Nav/Mobile/navMobile.css
```diff
@@ -78,8 +78,9 @@
   }
   .container-search-results {
     height: 50vh !important;
-    width: 100% !important;
+    width: 82vw !important;
     position: inherit !important;
```
mruwnik, line 82: is it possible to do this without the `important`s? They're in general toxic, so should be used as sparingly as possible

mruwnik (approved):

app/components/Nav/Mobile/navMobile.css
mruwnik: how about wrapping them in a media query, so they are only applied on mobile/desktop?

buddy-web3: made a new class for them on mobile, loading it last

PR #462, opened 2024-03-01
design requests

buddy-web3:


app/components/Nav/Mobile/navMobile.css
```diff
@@ -67,6 +67,7 @@
     align-content: center;
     justify-content: flex-start;
     align-items: flex-start;
+    width: 100%;
```
mruwnik, line 70: apply the `full-width` class

app/components/Nav/Mobile/navMobile.css
```diff
@@ -78,9 +79,12 @@
...
   .container-search-results-mobile {
     height: 50vh;
-    width: 82vw;
+    width: 100%;
     position: inherit;
     margin-top: var(--spacing-8);
     overflow: scroll;
   }
+  .mobile-searchbar > div {
+    width: 100%;
```
mruwnik, line 88: ditto

app/components/SearchInput/Input.tsx
```diff
@@ -19,14 +19,16 @@ interface SearchInputProps {
 }
 export const SearchInput = ({onChange, expandable, placeholderText}: SearchInputProps) => {
   const [search, setSearch] = useState('')
-
+  const inputRef = useRef<HTMLInputElement | null>(null)
```
mruwnik, line 22: try `const inputRef = useRef<HTMLInputElement>()` - undefined is better for this kind of thing

app/components/SearchInput/Input.tsx
```diff
@@ -44,6 +46,7 @@ export const SearchInput = ({onChange, expandable, placeholderText}: SearchInput
             handleSearch(e.currentTarget.value)
           }}
           value={search}
+          ref={inputRef}
```
mruwnik, line 49: have you tried just using the `autoFocus` attribute?

app/components/SearchResults/Dropdown.tsx
```diff
@@ -37,7 +39,9 @@ export const SearchResults = ({results}: {results: SearchResultsProps[]}) => {
         <Link key={i} className="search-result" to={result.url}>
           <Paper />
           <div className="search-result-content">
-            <h5 className="search-result-title">{result.title}</h5>
+            <h5 className={['search-result-title', isMobile ? 'xs-bold' : ''].join(' ')}>
```
mruwnik, line 42: can you do this with a media query, rather than adding clases?

app/components/SearchInput/Input.tsx
buddy-web3: that's a lint error

app/components/SearchResults/Dropdown.tsx
buddy-web3: How do I add a class to another class with a media query?

mruwnik: you don't :/ This way is fine, but it would be nice avoid explicitly checking the screen size in js, if possible

buddy-web3: To resolve your first 2 comments, I would need to put 1 js screen size check for the first one.
And for the second one, a check large if loop. 
Should I still resolve the first 2?

mruwnik: which comments? I'm not sure I understand

buddy-web3: The 2 regarding `full-width` class

mruwnik: wait, so the items in navBar.css aren't for that component? Why aren't they in the appropriate css file? 
But no point in adding the classes, then :D

buddy-web3: Those are for the search results. Same-same, maybe slightly different, but same:)

melissasamworth (approved):

PR #464, opened 2024-03-01
fix see more

mruwnik:


app/server-utils/stampy.ts
```diff
@@ -260,8 +260,17 @@ const renderText = (pageid: PageId, markdown: string | null): string | null => {
...
 
+// Icons can be returned as strings or objects
+const extractIcon = (val?: string | string[] | Entity | Entity[]): string | undefined => {
+  if (!val) return val
+  const item = head(val)
+  if (typeof item === 'string') return extractText(val as string)
+  if (item) return (item as Entity).url
+  return undefined
+}
+
```
jrhender, line 271: I think it would be nice to avoid the `val as` casts to benefit from type checking of `val` in the future. I've suggested an `isStringOrStringArray` type guard to do this, though there might be a more elegant way.

```suggestion
  if (isStringOrStringArray(val)) return extractText(val)
  const item = head(val)
  if (item) return item.url
}

const isStringOrStringArray = (x: unknown): x is string | string[] => {
  const item = head(x)
  return typeof item === 'string'
}

```

I also think the last `return undefined` isn't necessary. I think `undefined` will be returned if none of the `if` cases are matched without this line.

jrhender (commented):
A few suggestions

app/server-utils/stampy.ts
```diff
@@ -132,7 +132,7 @@ export type AnswersRow = CodaRowCommon & {
     Banners: '' | Entity[]
     'Rich Text': string
     Subtitle?: string
-    Icon?: string
+    Icon?: Entity | string | Entity[]
```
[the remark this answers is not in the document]
mruwnik: this is here mainly to document what Coda returns. Which can be either the raw image (base64 encoded), an url to the image, a reference to a Coda item or a list of references to Coda items... I'm not sure what exactly causes what

app/server-utils/stampy.ts
mruwnik: the thing is that I know that if it's not a string and not undefined, then it's going to be an `Entity` at this point. Typescript doesn't know that.

Wouldn't that be `void`?

jrhender: > the thing is that I know that if it's not a string and not undefined, then it's going to be an Entity at this point. Typescript doesn't know that.

Ah maybe part of the confusion here is that in my local code where I was trying this out, I also changed the signature of the `head` function to `const head = <T,>(item: T | T[]): T`. With this, I'm pretty sure that typescript does know that `item` is an `Entity` at that point. You can see in screenshot below that my local compiler is showing the type of `item` as that point as `Entity`. 

image

The way I'm reasoning about it is that the parameter type is `undefined | string | string[] | Entity | Entity[]` and then:
1. `if (!val) return val` removes the possibility of `undefined`
2. `if (isStringOrStringArray(val)) return extractText(val)` removes the possibility of `string | string[]`
3. `head` (with the signature change above) narrows `Entity | Entity[]` to `Entity`

At any rate, I'll make a PR at some point to improve the `head` function so that it's easy to make future code more type safe 😊 . Probably less important getting the UI redesign out.

> Wouldn't that be void?
No, I don't think so. I think `void` is a typescript type but the actual return type is `undefined`. I wrote this JS fiddle as a quick test: https://jsfiddle.net/#&togetherjs=M9fHnDlskK

mruwnik: ohhh, I like that version of head!

PR #501, opened 2024-03-07
QA requests

buddy-web3:


app/newRoot.css
```diff
@@ -529,3 +529,10 @@ ol {
     flex-direction: column;
   }
 } /* end mobile */
+p,
+button {
+  font-family: Poppins;
+}
+.margin-top-40 {
```
melissasamworth, line 536: we have a padding top class. Do you need margin specifically?

app/newRoot.css
```diff
@@ -529,3 +529,10 @@ ol {
     flex-direction: column;
   }
 } /* end mobile */
+p,
+button {
+  font-family: Poppins;
```
melissasamworth, line 534: this is specified in body, why does it need to be specified again?

melissasamworth (changes requested):

app/newRoot.css
buddy-web3: yes

app/newRoot.css
buddy-web3: yes. that's why buttons still are Arial

melissasamworth (approved):

PR #543, opened 2024-03-08
Stampy redesign

mruwnik:
QA here: https://redesign.aprillion.workers.dev


melissasamworth (changes requested):
I have two launch-blocking issues
image

README.md
Aprillion, line 1: @mruwnik / @buddy-web3  could you update README, please? this was not the only change, right? 🤔

Aprillion (commented):
...I checked a few files and I don't have any big objections. The site after redesign was supposed to be more professional than the old one, so I am not sure about releasing before fixing most of the small bugs too and cleaning up code for unused features or whatever is still duplicated, and updating README and/or other documentation ... but I will leave that up the core team. I can help after 2024-03-30 to do more polishing even after release I guess.

FTR, https://www.aisafety.com/reading-guide is now updated to use iframe from `stampy.ai` and DNS in Cloudflare is updated to point stampy.ai to https://stampy-ui.aprillion.workers.dev/ => migration of embed version is not a blocker any more.

melissasamworth (approved):

Issue #531, opened 2024-03-07
/articles mobile: secondary nav text colour + chevron cut off


ricewithabee:
Site:

![Image](https://github.com/StampyAI/stampy-ui/assets/98026499/5ea4a281-87e5-484f-b3c0-2817be9cf68d)

Figma:

![Image](https://github.com/StampyAI/stampy-ui/assets/98026499/94ea0b7e-3283-45d1-9c4c-807365da6064)

jrhender:
Reproduced again on aisafety.info at dd1e10e and locally at [8b67900](https://github.com/StampyAI/stampy-ui/pull/574/commits/8b67900bf992f07fd80fa12fc14e9f6c79743444) using chrome and safari on iPhone.
Perhaps this is a webkit mobile browser issue again.

jrhender:
Seems like the style is coming the "user agent stylesheet", so my thinking is that this button just needs a colour class.

![Image](https://github.com/StampyAI/stampy-ui/assets/4407464/ba4e6517-19a9-4265-bad9-30963c61afd1)

PR #577, opened 2024-04-01
fix(questions route): add black class to mobile nav button

jrhender:
Fixes: #531 

From the figma, I think this colour should be `cool grey 900`
image

But from the css, `cool-grey-900` is defined [to be the same as black](https://github.com/StampyAI/stampy-ui/blob/master/app/newRoot.css#L6). And as [we have a class for black](https://github.com/StampyAI/stampy-ui/blob/dd1e10e185326fcbffbb4b9ef4abfbc3944bd4da/app/newRoot.css#L164) but no class for cool grey 900, I figured I would go ahead with adding the `black` class.

I tested fix on safari on Iphone (where I was able to reproduce this error previously).

### DIFF
--- a/app/routes/questions.$questionId.$.tsx
+++ b/app/routes/questions.$questionId.$.tsx
@@ -119,7 +119,7 @@ export default function RenderArticle() {
           </div>
         ) : (
           <Button
-            className="mobile-only article-selector large-reading"
+            className="mobile-only article-selector large-reading black"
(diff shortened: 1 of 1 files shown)


melissasamworth:
@jrhender thanks! I think we do have a class for both, though. The engineers added black because they were confused by cool grey 900 so I let it fly lol.

melissasamworth (approved):

PR #582, opened 2024-04-08
remove Feedback popup

buddy-web3:


melissasamworth (changes requested):
we want to bring the feedback form back eventually, so it's prob best to comment it out instead of deleting the code

buddy-web3:
@melissasamworth it can always be brought back from Version History.

melissasamworth (approved):

PR #601, opened 2024-04-29
use human answers when available

mruwnik:


app/components/Chatbot/chat_entry.css
```diff
@@ -30,10 +30,66 @@ article.stampy {
...
+}
+
+.ref-6 {
+  background: rgb(231, 255, 178);
+  color: rgb(99, 159, 4);
+}
+
+.ref-7 {
+  background: rgb(231, 232, 255);
+  color: rgb(77, 75, 254);
```
melissasamworth, line 77: curious why you use rgb instead of hex

melissasamworth (approved):

Issue #591, opened 2024-04-26
Chatbot: Question input


melissasamworth:
**Acceptance criteria**

CSS stuff (Archana or Melissa)
- The questions input matches the visuals of the design (linked)
- Fixed to the bottom of the page all the time, even when scrolling (see Chatgpt)
- Make sure the area below the input is a white bar so that when there are lots of answers and the user scrolls, we can't see the answers show up there (see chatgpt)

Other stuff (Nemo or Buddy)
- The questions input shows on every page
- Before a user asks a question, the placeholder text reads "Ask Stampy a question..."
- After the user asks a question, the placeholder text reads "Message Stampy"

**Links**

https://www.figma.com/file/qqfeYAh6Z7QdyJXsQjIUfh/Stampy?type=design&node-id=1156%3A19217&mode=design&t=dg1esxPyfTqHQMnF-1

melissasamworth:
Update: still styling issues here too. We can move it over once we address them, mostly likely Bryce has found them

melissasamworth:
Update: Melissa's video with some small feedback

Issue #590, opened 2024-04-26
Chatbot: prototype warning


melissasamworth:
**Acceptance criteria**

Front-end

- The warning is in the bottom right and stays in the bottom right, even when user scrolls
- The visuals match the design (linked)

**Links**
https://www.figma.com/file/qqfeYAh6Z7QdyJXsQjIUfh/Stampy?type=design&node-id=1156%3A19217&mode=design&t=dg1esxPyfTqHQMnF-1

PR #602, opened 2024-04-30
Stampy redesign 591

buddy-web3:
#591 #590


app/components/Chatbot/index.tsx
```diff
@@ -248,7 +249,7 @@ export const Chatbot = ({question, questions}: {question?: string; questions?: s
   })
 
   return (
-    <div className="centered col-9 padding-bottom-128">
+    <div className="centered col-9 height-70">
```
melissasamworth, line 252: I believe this should take up 10 columns
image

melissasamworth: If you want to leave this to me when I do css that's totally fine

app/components/Chatbot/widgit.css
```diff
@@ -20,3 +20,29 @@
     width: 100%;
   }
 }
+.height-70 {
```
melissasamworth, line 23: oh interesting. For now on we could do .height-70vh so we know the units!

app/components/Chatbot/widgit.css
```diff
@@ -20,3 +20,29 @@
...
+}
+.warning-floating {
+  position: fixed;
+  right: 7vw;
+  z-index: 100;
+  bottom: 5vw;
+  width: 11vw;
+}
+.red {
+  color: #d40000;
```
melissasamworth, line 47: We'll want to define this and reference it as a variable, plus put this class in the same place as the other color classes in root.css

app/components/Nav/index.tsx
```diff
@@ -28,6 +29,12 @@ export const Nav = ({toc, categories}: NavProps) => {
             id="showArticles"
           />
           <ArticlesDropdown toc={toc} categories={categories || []} />
+          <MenuItem
+            primary={true}
+            link="https://chat.aisafety.info"
+            icon={<BotIcon />}
+            text="Stampy chatbot"
```
melissasamworth, line 36: This is the global nav right? Could we make it say "AI Safety chatbot" instead? :)

melissasamworth (changes requested):

buddy-web3:
@melissasamworth "If you want to leave this to me when I do css that's totally fine" - yes please, take over the design parts. Please note that all of those are unrelated to the tickets.. they are some design changes that I had stashed and now somehow got pushed.

melissasamworth (approved):

PR #603, opened 2024-04-30
Chat settings

mruwnik:


app/components/Chatbot/widgit.css
```diff
@@ -20,3 +20,16 @@
     width: 100%;
   }
 }
+
+.settings-container {
+  position: absolute;
+  bottom: var(--spacing-16);
+  left: var(--spacing-16);
+}
+
```
melissasamworth, line 29: Aw thanks. Feel free to do the minimum when it comes to CSS

melissasamworth (approved):

app/components/Chatbot/widgit.css
mruwnik: oh, I will :D

PR #610, opened 2024-05-02
reintroduce vw-based .col-x

Aprillion:
* in the spirit of previous `.col-x` classes with 12 columns, but should be compatible with `body {max-width: ...}` now
* https://stampy-ui2.aprillion.workers.dev/

### DIFF
--- a/app/newRoot.css
+++ b/app/newRoot.css
@@ -73,7 +73,7 @@ input {
 }
 body {
   margin: auto;
-  max-width: 1600px;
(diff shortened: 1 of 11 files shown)


app/newRoot.css
melissasamworth, line 1: The concept is right, but I'm going to go in and change some of the values as these don't account for margins (e.g. 12 columns shouldn't be full width)

melissasamworth (dismissed):

melissasamworth:
Whoops just realized this should not have gone to master! Should have gone to stampy-redesign

PR #619, opened 2024-05-03
suggestions in chatbot

mruwnik:


app/root.css
```diff
@@ -291,51 +291,51 @@ h2 {
 /* note I may change these to vars later */
 
 .col-1 {
-  width: clamp(0, 3.07vw, 53px);
+  width: clamp(0px, 3.07vw, 53px);
```
Aprillion, line 294: oh, it's totally evil that unitless 0 is not supported in clamp 😱

Aprillion (approved):

app/root.css
mruwnik: yes!!! Took me far too long to work that out...

Issue #592, opened 2024-04-26
Chatbot: question-matching on questions input


melissasamworth:
**Acceptance criteria**

CSS (Melissa, Archana)
- Make sure visuals of input match design (including hover state)

Other stuff (Nemo, Buddy)
- As soon as the user starts typing, the most closely matching answer in our database shows above it for user to click
- I imagine about 1 second (or more?) is good to return the match, because we need maybe like half a sentence before we can really match it (can spec out more based on what people think)

**Links**
https://www.figma.com/file/qqfeYAh6Z7QdyJXsQjIUfh/Stampy?type=design&node-id=1156%3A19240&mode=design&t=dg1esxPyfTqHQMnF-1

melissasamworth:
I _think_ you are done this one, just want to confirm

melissasamworth:
@varchanaiyer do you have any interest in taking the visuals for this one? The text in the little thing that shows above with a question should be bigger (.default I think) to match the input text size. Also ideally a hover state, the bg becomes super light teal (remember to use vars and reference the pre-existing type classes!)

melissasamworth:
Update: visuals should be good to go, pr of mine pending

@mruwnik there seems to be a significant lag in some places, e.g. type in "What is AI Safety" and it takes a few seconds to show up when otherwise the answers show instantly

melissasamworth:
Going through again.

This is not a huge deal so if it takes a long time don't do it.

But the logic for choosing a question seems not right. For instance

"What is" produces
![Image](https://github.com/StampyAI/stampy-ui/assets/825910/f402174f-e0b8-48bf-98f5-bec6d5744e9d)

But "What is " with  a space produces

![Image](https://github.com/StampyAI/stampy-ui/assets/825910/21cee1f1-f08f-42f6-922e-5916709877b5)

"What is AI safet" produces nothing but "What is AI safety" shows

![Image](https://github.com/StampyAI/stampy-ui/assets/825910/e60ed557-427b-4ac7-80ee-a9a835741140)

Ideally if there is a match to a question e.g. "Do people seriously worry about existential risk from AI?" then every version of this that is less characters e.g. "Do people seriously" will also show something. It wouldn't appear then disappear then show again.

mruwnik:
the last bit is more complicated, as it changes how search works - I propose creating a separate task for that

PR #629, opened 2024-05-09
Chatbot: question-matching on questions input #592

melissasamworth:


mruwnik (approved):

melissasamworth:
Merged this even though the check had failed. I don't know what the consequence of this is. I probably shouldn't have done it but I was getting impatient.

mruwnik:
you shouldn't have, true... The check looks like it's not your fault though, so it would have to be fixed either way

PR #630, opened 2024-05-09
inform Feedback type. Submit positive feedback

buddy-web3:


app/components/Feedback/index.tsx
```diff
@@ -57,7 +57,10 @@ const Feedback = ({
           showText={!!labels}
           actionType={ActionType.HELPFUL}
           hint={upHint}
-          onClick={() => setShowFeedback(true)}
+          onClick={() => {
+            setShowFeedback(true)
+            onSubmit && onSubmit('', '')
```
mruwnik, line 62: why not just send in a `Thumbs up` feedback or something as the option?

PR #631, opened 2024-05-09
clean up fcol-x lefovers from merge

Aprillion:
https://stampy-ui2.aprillion.workers.dev/

### DIFF
--- a/app/root.css
+++ b/app/root.css
@@ -43,7 +43,7 @@
   --spacing-104: clamp(56px, 10.6vw, 104px);
   --spacing-128: clamp(56px, 13.8vw, 128px);
   --spacing-192: clamp(80px, 17vw, 192px);
-  --spacing-288: clamp(128px, 25.6vw, 288px);
+  --spacing-288: clamp(128px, 20vw, 288px);
 }
 
 /* CSS Reset */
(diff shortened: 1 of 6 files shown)


app/root.css
```diff
@@ -43,7 +43,7 @@
   --spacing-104: clamp(56px, 10.6vw, 104px);
   --spacing-128: clamp(56px, 13.8vw, 128px);
   --spacing-192: clamp(80px, 17vw, 192px);
-  --spacing-288: clamp(128px, 25.6vw, 288px);
+  --spacing-288: clamp(128px, 20vw, 288px);
```
melissasamworth, line 46: I'm still uncertain about making the spacing variables vw. I use a system based on 8 to make it clean. But let's try it and see.

I definitely think we should make the columns vw with clamp, which I believe is the case now!

app/root.css
```diff
@@ -498,7 +464,7 @@ svg {
 
 /* defines the standard left and right margins */
 .page-body {
-  padding: 0px 13.333vw;
+  padding: 0px var(--spacing-128);
```
melissasamworth, line 467: This makes me uncomfortable but I'll merge it anyway

melissasamworth (approved):

app/root.css
Aprillion: vw padding on clamped width means narrower content on larger screens... so the padding also needs to be clamped (though the "correct" value might need to be changed after design review on various screen sizes I guess .. I just picked a closest option to the 13vw)

PR #634, opened 2024-05-20
Stampy redesign 376

varchanaiyer:


mruwnik:
don't merge to master - the redesign branch is stampy-redesign

mruwnik:
there are some conflicts that need to be fixed here

app/components/Chatbot/ChatEntry.tsx
```diff
@@ -1,5 +1,5 @@
-import {ComponentType} from 'react'
-import {Link} from '@remix-run/react'
+import { ComponentType } from 'react'
```
melissasamworth, line 1: What was the purpose of adding all these spaces?

melissasamworth (commented):
Gah I am sorry but we cannot merge this! There are changes to the root.css file, which I am very protective of (I'm sorry, I should have expressed that I am really pedantic that no one touches the tags and one-off classes, particularly the typography classes).

Also there are too many conflicts. As a result I can't see what is changed. All I see is that we are missing the feedback widget under the human-written responses.

I would say I'm going to delete this request, and you can make a new one. So save the code you want to save! Hope that okay, I think saving it will just not be worth it at this point.

app/components/Chatbot/ChatEntry.tsx
mruwnik: probably auto import + the editor doing it. @varchanaiyer, make sure to run `npm run lint:fix` before pushing, as that will both make the code style consistent and notify you of any errors

Issue #598, opened 2024-04-27
Chatbot: feedback


melissasamworth:
**Acceptance criteria**

Everything except CSS (Nemo, Buddy)
- Show bar with icons under every Stampy response
  - Under bot generated responses, show just the thumbs up and thumbs down
- Hover states for icons is
  - A centered tooltip above (ideally it would show below based on scroll position but I think this is not "built in" for us, i.e. extra work, so skip that for now if it would take more than 1 minute)
  - A light grey bg underneath the icon user is hovering over

![Image](https://github.com/StampyAI/stampy-ui/assets/825910/2f696e5b-abca-4fb3-809f-a6a6add71328)

- If user clicks thumbs up
  - "Thank you for your feedback!" text appears to right
  - Thumbs up icon is now highlighted with light teal bg and icon itself is dark teal
  - User cannot change answer (just to make it easier for us)
- If user clicks thumbs down
  - Thumbs down icon is now highlighted with light teal bg and icon itself is dark teal
  - User cannot change answer (just to make it easier for us)
  - Feedback popover appears above and to the right
    - Submit feedback button is greyed out until user chooses one of the 5 categories
    - User can leave a comment but doesn't have to
    - User can click out of the modal to make it disappear (in this case the thumbs down icon stays highlighted and we'll still record this data)
    - If user clicks "Submit feedback" then "Thank you for your feedback!" text appears to right

CSS (Archana, Melissa)
- Make sure visuals match the design, including hover states

**Links**
https://www.figma.com/file/qqfeYAh6Z7QdyJXsQjIUfh/Stampy?type=design&node-id=2197%3A6785&mode=dev&t=Isg5hjkFzr1KY4KQ-1

melissasamworth:
update: only done for human answers

melissasamworth:
Update: Melissa still needs to adjust CSS

melissasamworth:
@mruwnik is this all hooked up to the Discord server? If not lmk and I can assign you a task for that

melissasamworth:
Update: I still have to look at Myles's pr

melissasamworth:
Update: It's still not letting me merge, messaged Myles about it. Once it's merged move it to design review!

melissasamworth:
Update: Melissa's video with some small feedback

PR #639, opened 2024-05-25
Ticket #598 updated

zarSou9:
Unlike the previous pr for this ticket, this one uses the `Action` component. I've created a `ButtonSecondary` component as a smaller alternative to the `Button` component with slightly different styling as to match the Figma.

`tailwindcss` was added to improve the styling experience

Note: All 3 imports in the root.css file are needed for tailwind to work correctly. Also turns out the reference problems were happening cus I was using pnpm.


app/components/Feedback/index.tsx
```diff
@@ -39,48 +40,69 @@ const Feedback = ({
...
+  const [vote, setVote]: any = useState(undefined)
+
+  const thanksRef = useRef<HTMLDivElement | null>(null)
 
   useEffect(() => {
-    const timeout = setInterval(() => setShowFeedback(false), 6000)
-    return () => clearInterval(timeout)
-  }, [showFeedback])
+    if (showThanks && thanksRef.current) {
+      thanksRef.current.style.opacity = '1'
```
mruwnik, line 54: use classes for css. Or contitional rendering if you want to use js. Don't mix the two - there lies madness

mruwnik: why this way, rather than the previous approach?

app/components/Feedback/index.tsx
```diff
@@ -39,48 +40,69 @@ const Feedback = ({
...
-    <div className="feedback relative">
-      <CompositeButton className="flex-container">
+    <div className="flex items-center">
+      <ButtonSecondaryWrapper>
         <Action
           pageid={pageid}
           showText={!!labels}
           actionType={ActionType.HELPFUL}
+          active={vote === 'helpful'}
+          dissabled={!!vote}
```
mruwnik, line 69: typo

app/components/Feedback/index.tsx
```diff
@@ -39,48 +40,69 @@ const Feedback = ({
   upHint,
   downHint,
   options,
+  formClassName,
   onSubmit,
 }: FeedbackProps) => {
-  const [showFeedback, setShowFeedback] = useState(false)
+  const [showThanks, setShowThanks] = useState(false)
   const [showFeedbackForm, setShowFeedbackForm] = useState(false)
+  const [vote, setVote]: any = useState(undefined)
```
mruwnik, line 48: this means that each time the page gets refreshed, the user can vote for the article again. The previous solution would rememeber whether an article was liked or not between refreshes

app/components/ButtonSecondary/button_secondary.css
```diff
@@ -0,0 +1,15 @@
+.button-secondary .tool-tip {
```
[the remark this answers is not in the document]
zarSou9: Tailwind does have some limitations. i.e. it is difficult to use selectors to control the styling of child elements.

app/routes/questions.actions.tsx
```diff
@@ -89,26 +97,26 @@ export const action = async ({request}: ActionFunctionArgs) => {
 type Props = {
   pageid: string
   actionType: ActionType
-  showText?: boolean
   hint?: string
-  children?: ReactNode | ReactNode[]
```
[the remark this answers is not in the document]
zarSou9: You right, I added it back

app/routes/questions.actions.tsx
```diff
@@ -89,26 +97,26 @@ export const action = async ({request}: ActionFunctionArgs) => {
...
+  dissabled?: boolean
```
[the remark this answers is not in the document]
zarSou9: fixed

app/components/Feedback/index.tsx
zarSou9: You right on both accounts. I've updated the code

app/components/Feedback/index.tsx
zarSou9: yup

app/components/Feedback/index.tsx
zarSou9: very good point. I've updated the code, but note that I had to pass a setVoted function to the child Action components so that the feedback knew to set the other Action disabled. Its the best I could do. Also the handler function for voting in the questions.actions file is throwing an error, and thus the feedback button is returning to its null state after you press it.

zarSou9:
I've incorporated all of your feedback except the check is failing for absolutely now reason. A prettier error even though I have run prettier. Oh well, I'm going to sleep

zarSou9:
Ok, no more tailwind, should be ready to merge 🤞

app/components/Article/article.css
```diff
@@ -154,17 +154,17 @@ article .link-popup p {
...
+.form {
+  left: 845px;
+}
+
 article .footer-comtainer {
   display: flex;
   align-items: center;
   margin-top: var(--spacing-56);
   margin-bottom: var(--spacing-24);
+  gap: 16px;
```
mruwnik, line 166: try to avoid raw pixel values. In `root.css` there's a bunch of available variables that should be used for sizing and coiours. You use them as `var(<var name)`, like in the previous lines. So in this case it'd be `gap: var(--spacing-16);`. This is important both for consistency (as it's easier to change everything at once), but also because the mobile version often has subtle differences in spacing which should be reflected by this

app/components/ButtonSecondary/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+        }
+        onClick={(e) => {
+          if (disabled) {
+            e.preventDefault()
+          }
+        }}
+        {...props}
+      >
+        {children}
+        {tooltip && !disabled && <p className="tool-tip">{tooltip}</p>}
```
mruwnik, line 40: from what I can see, this is the only major difference between this component and the `Button` one? That doesn't seem to merit a totally differenct component, as you can achieve the same by doing
```
<Button
   action={action}
   className={
          'button-secondary ' +
          (active ? 'active' : disabled ? '' : 'inactive') +
          (className ? ' ' + className : '')
        }
    active={active}
    disabled={disabled}
    {...props}>
       {children}
        {tooltip && !disabled && <p className="tool-tip">{tooltip}</p>}
</Button>
```
which in this case would be:
```
          <Button
            action={question.answerEditLink || ''}
            props={{target: '_blank', rel: 'noopener noreferrer'}}
        className="button-secondary"
          >
            <EditIcon className="no-fill" />
           <p className="tool-tip">Google Doc</p>
          </Button>
```

app/components/Feedback/feedback.css
```diff
@@ -1,23 +1,49 @@
...
+  width: 384px;
+  padding: 30px;
+  border-radius: 6px;
 }
 
 /* Comment to Nemo when he reviews: why didn't you just do this the way you did the settings buttons? */
 .select-option {
   height: var(--spacing-48);
-  padding: var(--spacing-8) var(--spacing-16);
+  border-radius: 6px;
```
mruwnik, line 33: use a `var` here. Or consider adding the `rounded` class to whatever is using this

app/components/Feedback/feedback.css
```diff
@@ -1,23 +1,49 @@
...
+  flex-direction: column;
+  position: fixed;
+  top: 50px;
+  left: calc(13.333vw + 85px);
+  width: 384px;
+  padding: 30px;
+  border-radius: 6px;
 }
 
 /* Comment to Nemo when he reviews: why didn't you just do this the way you did the settings buttons? */
```
mruwnik, line 30: check `git blame`. I didn't do this :D Feel free to make things more consistent everywhere

app/components/Feedback/feedback.css
```diff
@@ -1,23 +1,49 @@
...
-  padding: var(--spacing-8) var(--spacing-16);
+  border-radius: 6px;
   margin-bottom: var(--spacing-16);
   cursor: pointer;
   width: 100%;
 }
+
+.submit {
+  height: var(--spacing-48);
+  border-radius: 6px;
```
mruwnik, line 41: ditto

app/components/Feedback/feedback.css
```diff
@@ -1,23 +1,49 @@
...
+  font-weight: 500 !important;
```
mruwnik, line 42: @melissasamworth watch out, here she comes

app/components/Feedback/feedback.css
```diff
@@ -28,7 +54,6 @@
   padding: var(--spacing-12);
   margin: var(--spacing-40) 0;
   box-sizing: content-box;
-  background: pink;
```
mruwnik, line 31: hmm. interesting...

app/components/Feedback/index.tsx
```diff
@@ -39,48 +40,59 @@ const Feedback = ({
...
-            setShowFeedback(!showForm)
+            if (!showForm) setShowThanks(true)
             setShowFeedbackForm(!!showForm)
           }}
         />
-      </CompositeButton>
+      </ButtonSecondaryWrapper>
 
-      {showFeedback && <div className="thanks">Thanks for your feedback!</div>}
+      <p className={'thanks ' + (showThanks ? 'show' : 'hide')}>Thank you for your feedback!</p>
```
mruwnik, line 83: out of curiosity, why are you doing it this way?

app/components/Feedback/index.tsx
```diff
@@ -39,48 +40,59 @@ const Feedback = ({
...
 
       {showFeedbackForm && (
         <FeedbackForm
-          onSubmit={onSubmit}
+          onSubmit={(att) => {
```
mruwnik, line 87: can you name this something more intuitive?

app/root.css
```diff
@@ -160,6 +160,9 @@ h2 {
   letter-spacing: -0.14px;
 }
 
+.leading-0 {
+  line-height: 0 !important;
```
mruwnik, line 164: @melissasamworth

app/routes/questions.actions.tsx
```diff
@@ -173,13 +184,9 @@ export const Action = ({
       <input type="hidden" name="incBy" value={actionTaken ? -1 : 1} />
       <input type="hidden" name="stateString" value={stateString} />
       {children}
-      <Button className={className}>
+      <ButtonSecondary disabled={disabled} active={actionTaken} tooltip={hint}>
```
mruwnik, line 187: this is also just a normal button with a custom tooltip

app/components/Article/article.css
zarSou9: Agreed

app/components/ButtonSecondary/index.tsx
zarSou9: Yes yes, you are correct. I've changed it, but note that I decided to put the tool tip element within the button component for consistency.

app/components/Feedback/feedback.css
zarSou9: yes I've changed it. The only reason I used the 6px is because that's what the figma said, but the var for --border-radius is 4px. I totally get the value of only needing to change one variable though so yeah.

app/components/Feedback/index.tsx
zarSou9: Too true

app/routes/questions.actions.tsx
zarSou9: yup

app/components/Button/button.css
```diff
@@ -20,6 +20,15 @@
   letter-spacing: -0.16px;
 }
 
+.button-secondary {
```
mruwnik, line 23: how does this differ from the `.secondary` class below? That is conceptually what's the difference?

mruwnik (commented):
@melissasamworth can you check the css bits? The code seems fine

app/root.css
zarSou9: When I was trying to style the button, It kept putting this annoying space below it. After some testing, I discovered it was a line-height issue. I've added this to the root incase there are other circumstances where there is a line-height problem.

app/components/Button/button.css
zarSou9: The second version of the button, as represented in the figma, is significantly stylistically different, mainly as a result of the difference in hover states (the main button shows and outline, and the second shows a background). This also effects how sibling buttons are grouped. The `.secondary` class is a slight difference to the main button version, and the `.button-secondary` class is a second version of the button.

app/components/Button/button.css
```diff
@@ -154,6 +163,22 @@
...
+.button-secondary:hover .tool-tip-secondary {
+  opacity: 1;
+}
+
+.active path {
+  stroke: var(--colors-teal-600);
+}
+
+.active {
+  background-color: #edfaf9;
```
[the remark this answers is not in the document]
zarSou9: you right

app/root.css
melissasamworth: Fine with me

melissasamworth: Except sometimes the button actually goes to two lines. In which case this would end up looking really weird and bad obviously.

app/components/Feedback/feedback.css
melissasamworth: There's actually a variable for border radius so use that var!

app/components/Feedback/feedback.css
melissasamworth: Heh, I think I was referencing something else which I now forget

melissasamworth: But Myles could you use a variable instead of padding: 30, that way we'll always stay using my system of multiples of 8. That said I imagine it's better to just set the height in this case. But yeah just change it to a var and we're good for now

melissasamworth (changes requested):

zarSou9 (commented):
testing review

melissasamworth (approved):

PR #683, opened 2024-06-04
Ticket #591

zarSou9:
Added white space under input, and fixed positioning i.e. the chat input is relative to the viewport for both the chatbot view and main page

### DIFF
--- a/app/components/Chatbot/widgit.css
+++ b/app/components/Chatbot/widgit.css
@@ -18,7 +18,6 @@
 
 .widget-ask {
   position: relative;
-  
 }
 .widget-ask.fixed {
   position: fixed;
--- a/app/components/Chatbot/widgit.css
+++ b/app/components/Chatbot/widgit.css
@@ -17,8 +17,8 @@
 }
 
 .widget-ask {
-  position: sticky;
-  bottom: 0;
+  position: fixed;
+  bottom: 10px;
 }
 .right-icon {
   padding-right: var(--spacing-56);
--- a/app/components/Chatbot/widgit.css
+++ b/app/components/Chatbot/widgit.css
@@ -17,9 +17,13 @@
 }
 
 .widget-ask {
+  position: relative;
+}
+.widget-ask.fixed {
   position: fixed;
   bottom: 50px;
 }
+
 .right-icon {
   padding-right: var(--spacing-56);
 }
--- a/app/components/Chatbot/widgit.css
+++ b/app/components/Chatbot/widgit.css
@@ -18,7 +18,7 @@
 
 .widget-ask {
   position: fixed;
-  bottom: 10px;
+  bottom: 50px;
 }
 .right-icon {
   padding-right: var(--spacing-56);
@@ -63,4 +63,14 @@
   position: absolute;
   right: 4px;
   bottom: 4px;
+}
+
+.white-space {
+  position: absolute;
+  left: -1000px;
+  width: calc(100% + 2000px);
+  height: 100px;
+  top: 100%;
+  z-index: 1000;
+  background: #ffffff;
 }
--- a/app/components/Chatbot/widgit.css
+++ b/app/components/Chatbot/widgit.css
@@ -21,7 +21,7 @@
 }
 .widget-ask.fixed {
   position: fixed;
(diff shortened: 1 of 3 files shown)


app/components/Chatbot/widgit.css
```diff
@@ -17,9 +17,13 @@
 }
 
 .widget-ask {
-  position: sticky;
-  bottom: 0;
+  position: relative;
 }
+.widget-ask.fixed {
+  position: fixed;
+  bottom: 50px;
```
melissasamworth, line 24: Use a var!

app/components/Input/input.css
```diff
@@ -6,10 +6,10 @@
   border-radius: var(--border-radius);
   border: 1px solid var(--colors-cool-grey-200, #dfe3e9);
   background: var(--colors-white, #fff);
-  box-shadow: 0px 16px 40px 0px rgba(175, 183, 194, 0.2);
   color: var(--colors-cool-grey-900, #1b2b3e);
   padding: var(--spacing-8) 0 var(--spacing-8) var(--spacing-12);
-  height: var(--spacing-48);
+  /* This can't be a variable because it shouldn't shrink when the screen gets small */
+  height: 48px;
```
melissasamworth, line 12: Yeah that's true. I think I may want to create height vars soon

app/components/Input/input.css
```diff
@@ -24,11 +24,15 @@
...
 
 .input.large {
-  height: var(--spacing-56);
+  height: 56px;
   font-size: 18px;
   letter-spacing: -0.18px;
 }
 
+.input.shadow {
+  box-shadow: 0px 16px 40px 0px rgba(175, 183, 194, 0.2);
```
melissasamworth, line 33: We can keep this for now, but general note is I wonder... If there's a way to just reference the same CSS that we already created for shadows. Like by referencing that class in the component itself. I.e. all input components have the class "shadow" (or whatever it's called) that we already defined. Ideally we would have a system where CSS never has to be written twice, to keep things perfectly consistent! 

@mruwnik could you comment on the feasibility of this?

app/components/Input/input.css
```diff
@@ -24,11 +24,15 @@
 /* size */
 
 .input.large {
-  height: var(--spacing-56);
+  height: 56px;
   font-size: 18px;
```
melissasamworth, line 28: Is it possible to assign the class .large to all components instead of re-writing the typography?

@mruwnik maybe chime in here?

zarSou9: Yes, you can just create the .large class in the root.css file and use it universally, but I think you already knew that

app/components/Input/input.css
zarSou9: I mean you can of course create a .shadow class in the root.css file and reference it to any component if that is what you mean. I just haven't been doing that because you said to not mess with the root.css file

app/components/Chatbot/widgit.css
zarSou9: you right, one sec

app/components/Input/input.css
mruwnik: I think her question was more about adding the `large` class to all components :D 
But yes, it's certainly possible - as long as you specifiy which ones

app/components/Input/input.css
mruwnik: I added a `shadowed` class to root.css that does this

mruwnik (commented):
LGTM - up to Melissa, though

app/components/Input/input.css
melissasamworth: Yes my question is about adding the large class to the components themselves, ideally we would use the typography classes I wrote that once and never write a line of typography again!

melissasamworth (approved):

app/components/Input/input.css
melissasamworth: I'm sorry, amendment: you can add your own classes to root.css, particularly one-off classes. Just don't add to or edit the classes I already made heh.

melissasamworth: But yes from now on reference Nemo's shadowed class!

PR #693, opened 2024-06-06
🦄

melissasamworth:


app/components/Chatbot/chat_entry.css
```diff
@@ -122,10 +122,11 @@ article.stampy {
   word-wrap: break-word;
   background-color: white;
   border: 1px solid var(--colors-cool-grey-200, #dfe3e9);
-  padding: var(--spacing-24) var(--spacing-32);
+  padding: var(--spacing-40);
   position: absolute;
   transform: translateX(50%);
   text-decoration: unset;
+  z-index: 3;
```
mruwnik, line 129: shouldn't the element just have the `z-index-3` class?

mruwnik (approved):

app/components/Chatbot/chat_entry.css
melissasamworth: Yeah probably

Issue #376, opened 2024-02-07
Chatbot: human-written responses


melissasamworth:
**Acceptance criteria** 

Everything besides CSS (Nemo, Buddy)

- Human-written responses are returned if semantic similarity from organic question exceeds a certain amount
  - Also when the user has clicked a canned question
  - Otherwise bot-generated response is returned
- Human-written responses
  - Are labeled "Human-written responses" in tag at top right of response
  - Have a grey background
  - Have links, images, styled text, and everything that articles have (the exact same as articles just narrower and with a grey bg)

CSS (Melissa, Archana)
- Make sure tag and responses match design

**Links**
https://www.figma.com/file/qqfeYAh6Z7QdyJXsQjIUfh/Stampy?type=design&node-id=1156%3A19268&mode=dev&t=9lrQdhpZsKoF1qGs-1

melissasamworth:
@mruwnik human written responses currently never show, it's always just a bot still!

varchanaiyer:
Feedback from Nemo:

```
the problem with that first click is that it needs to know what articles are available. So the way it works is:
- if the question asked (or clicked) is a known article, then just show that
- if the question is similar to a known article (0.9 similarity, I think), then show that
- otherwise, just call the chatbot

The issue here is that the second step only works when there is a set of known articles available (option 1 uses direct page ids, so is fine).
When the page is loaded, it fetches a list of article titles as well as an embedding matrix. For smaller queries (<= 2 words) or until the embedding matrix has been downloaded, articles are searched with a regex. For longer phrases (assuming the embeddings are available), the embedding is used. This works fine, as long as the embedding and/or list of article titles are available. The problem with this is that it can take a while for them to download. So if you open the chatbot, for the first second or two, you won't have search initialised. Which is why when you click on one of the suggested questions, the chatbot starts running
This isn't a problem with the followup questions, as they use the first mechanism, and contain ids to the articles they're pointing at, which allows them to be direct links
So the problem here is that Coda takes too long to respond
2 things that come to mind are checking whether the response from Coda is compressed, and limiting the amount of data returned from Coda - it's possible that it also returns the actual article contents along with the `titles`
```

This means that we need to figure out a way to improve the load time from coda. Potentially do some kind of pre-fetch?

varchanaiyer:
https://github.com/StampyAI/stampy-ui/pull/634

Human Written Responses UI changes done and responses turn after the first quuestion

melissasamworth:
[...]

PR #694, opened 2024-06-06
Ticket #376

zarSou9:
I've improved the hint styling for human-written responses, but not the other part of this ticket which was about improving the padding/spacing for articles. From my understanding, the styling for articles is done through some magic which converts google docs pages into well styled html. I think the way to fix spacing here is just to add/remove spaces in the docs


zarSou9:
@melissasamworth The article spacing is fixed for the chatbot. Should be ready to merge

melissasamworth (approved):

Issue #597, opened 2024-04-27
Chatbot: reference hover


melissasamworth:
**Acceptance criteria**

Everything except CSS (Nemo, Buddy)
- If user hovers over one of the little numbers, the referenced excerpt appears as a modal while the user is hovering
- The modal includes
  - Title of thing
  - Author
  - Type (e.g. blog post)
    - This links out in a new tab (not sure this makes sense given the modal will disappear on hover?)
  - The referenced excerpt
    - I'm not sure what other details to include here such as how much of the referenced excerpt to show, but I imagine just keep doing what you've already been doing

CSS (Archana, Melissa)
- Generally just make it look like the design yay
- Position the excerpt dynamically so that at least most of it is showing
  - If it is closer to the bottom of user's viewing window, position it directly above the reference number
  - If it is closer to the top of user's viewing window, position it directly below the reference number

**Links**
https://www.figma.com/file/qqfeYAh6Z7QdyJXsQjIUfh/Stampy?type=design&node-id=1865%3A4431&mode=dev&t=Isg5hjkFzr1KY4KQ-1

melissasamworth:
update: in pr

melissasamworth:
Update :I can't get a bot generated response to check but I imagine there are still issues. Hopefully  Bryce has found them

PR #701, opened 2024-06-07
Ticket-#597 - Reference Hover

zarSou9:
Ok so I think I've done it, except its a little bit janky. i.e. in order to get the height consistent and correct, instead of editing the dangerouslyPastedHTML I've just set the overflow: hidden, and set the height of the contents div to be one which generally doesn't seem to clip words. 

also, @melissasamworth I made a design decision to put the reference popup below the number (instead of above, which is on the figma), because I feel like it generally clips off less often because I notice I am more likely to be reading stuff at the top of the screen. I also like that the curser is closer to the title and link when its below. Anywho, I can change it if you disagree.

https://github.com/StampyAI/stampy-ui/assets/148247390/c8f3070f-706e-4456-97cf-4ab6180a28bb

### DIFF
--- a/app/components/Chatbot/index.tsx
+++ b/app/components/Chatbot/index.tsx
@@ -78,7 +78,10 @@ const QuestionInput = ({
   }
 
   return (
-    <div className={'widget-ask ' + (fixed ? 'fixed col-10' : 'col-9')} ref={clickDetectorRef}>
+    <div
+      className={'widget-ask ' + (fixed ? 'fixed col-10 z-index-4' : 'col-9')}
+      ref={clickDetectorRef}
+    >
       {results.length > 0 ? (
         <Button className="full-width suggestion" action={() => handleAsk(results[0].title)}>
           <p className="default">{results[0].title}</p>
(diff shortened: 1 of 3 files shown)


app/components/Chatbot/index.tsx
```diff
@@ -78,7 +78,10 @@ const QuestionInput = ({
   }
 
   return (
-    <div className={'widget-ask ' + (fixed ? 'fixed col-10' : 'col-9')} ref={clickDetectorRef}>
+    <div
+      className={'widget-ask ' + (fixed ? 'fixed col-10 z-index-4' : 'col-9')}
```
mruwnik, line 82: would `z-index-1` suffice? i.e. use the smallest possible value

melissasamworth (approved):
Looks amazing!

- I think your design decision makes sense.
- Bonus in the future is that it goes above or below based on user's scroll position
- Now for the height, I was just envisioning that we didn't have a static height, and the height was just as long as it needed to be. I think I would prefer that but I'm cool to use this and then test it as users and see how we feel.

PR #709, opened 2024-06-08
search by pageid instead of question

zarSou9:
Instead of messing with `onQuestion`, I decided to just use your `showFollowup` function at times when the pageId is available. e.i. followups, splash screen, and suggestions (I've renamed it to `showArticleByID` as such)  It seems to work great and its way faster now.


mruwnik (approved):
nice!

PR #711, opened 2024-06-08
properly display popup content

mruwnik:


app/hooks/useOnSiteQuestions.ts
```diff
@@ -0,0 +1,44 @@
...
+  '6568', // What is the orthogonality thesis? debugger eval code:1:26
+  '7755', // How powerful would a superintelligence become? debugger eval code:1:26
+  '8E41', // Will AI be able to think faster than humans? debugger eval code:1:26
+  '9B85', // Isn't the real concern misuse? debugger eval code:1:26
+  '8V5J', // Are AIs conscious? debugger eval code:1:26
+  '8IHO', // What are the differences between a singularity, an intelligence explosion, and a hard takeoff? debugger eval code:1:26
+  '6306', // What is an intelligence explosion? debugger eval code:1:26
+  '5943', // How might AGI kill people? debugger eval code:1:26
+  '7748', // What is a "warning shot"?
+]
```
zarSou9, line 25: big fan

zarSou9:
Code looks fire, and seems to be working. But just to make sure @melissasamworth, you still don't want the `loading...` to be in a big box? If so, I think we should at least remove the feedback buttons while it is loading.
image

Also, Nemo, I got this response to a question, I'm guessing there are unwritten questions where there shouldn't be in coda.
image

zarSou9:
Oh wait, there is also a bug, clicking on the settings causes the questions to reload. It actually might be nice to also have a refresh questions button come to think of it

https://github.com/StampyAI/stampy-ui/assets/148247390/e003b534-9797-47bb-889e-8756b7bfa76f

mruwnik:
good catch with the bug - @melissasamworth what do you think about it?

zarSou9:
@mruwnik Let me know when you've fixed the bug and resolved conflicts so I can merge. Also my humble opinion is to keep the big loading box but up to you

mruwnik:
done

zarSou9 (approved):

PR #724, opened 2024-06-13
misc fixes

zarSou9:


app/components/Article/Contents.tsx
```diff
@@ -40,6 +45,7 @@ const addPopup = (e: HTMLElement, id: string, contents: string, mobile?: boolean
     const toggle = () => popup.classList.toggle('shown')
     popup.addEventListener('click', togglePopup(toggle, e))
     e.addEventListener('click', togglePopup(toggle, e))
+    popup.children[0].addEventListener('click', (e) => e.stopPropagation())
   }
 
   return popup
```
zarSou9, line 51: @mruwnik, ok so there is a problem with multiple popups being added to a reference item for definitions. I think specifically, because the id is generated randomly, there can still exist two popups associated with the same reference whenever the function which creates the random ids is run more than once. And because react tends to just run things randomly...

zarSou9: I tried fixing this myself but I couldn't figure it out

mruwnik: hmm. You're right :/ One way round this would be to pass in the id, rather than use a random one. So normal articles would just leave it empty, and chatbot ones would pass in their index in the history. 

Otherwise the id would have to be generated based on the contents somehow, which would work for chat responses (as most of the time they will be unique), but not for articles in the chatbot (unless they have some magic token added secretly, which is also ugly). In this case the randomId would just be a hash of the `html` argument or something

mruwnik (approved):

app/components/Article/Contents.tsx
zarSou9: I've tried again but I think this problem has far deeper and darker roots...
image
They are replicating inside each-other. Its like trying to fight hydra

PR #734, opened 2024-06-17
No more youtube authors, and autocomplete for search off.

zarSou9:


app/components/SearchInput/Input.tsx
```diff
@@ -33,6 +33,7 @@ export const SearchInput = ({onChange, expandable, placeholderText}: SearchInput
         <MagnifyingIcon className="iconsMagnifyingGlass" />
         <input
           name="searchbar"
+          autoComplete="off"
```
mruwnik, line 36: why?

zarSou9: Sorry, I should have explained. It stops this from happening.
image
I remembered Melissa saying she didn't like the autocomplete suggestions, because it covered up the search results. In hindsight though I don't actually think its that bad, because it does serve a purpose of remembering previous searches. What do you think?

mruwnik: better to have just one or the other

mruwnik (approved):

Issue #643, opened 2024-05-27
Sometimes gets stuck on "Sending query..."


bryceerobertson:
![Image](https://github.com/StampyAI/stampy-ui/assets/98026499/674f176a-2c55-4a52-a636-5235549a4b30)

melissasamworth:
Update: needs steps to reproduce @bryceerobertson

bryceerobertson:
Prompts causing it to happen:

- "Test" > "Test again" > "Tell me about superintelligence"
- "Tell me about superintelligence" > "Are you a bot" > "Are you buggy"
- "Tell me something" > "Tell me something" > "Tell me something"

Conclusion: It happens on the third prompt, seemingly always.

bryceerobertson:
Looks like it's happening on the 4th prompt now 

<img width="1728" alt="Screenshot 2024-06-05 at 18 15 45" src="https://github.com/StampyAI/stampy-ui/assets/98026499/fe77bde1-cd48-4279-ad54-8cbab47308b3">

bryceerobertson:
It happens for me with these prompts:

"What is AI Safety?" > "What is an AI (2)" > "What is a robot (3)"
"Say something interesting" > "Another one" > "Is AI safe?" > "How safe is it?"
[A bunch of human-written responses, which don't seem to trigger the problem] > "How will we know" > "Is it even possible to know?"
<img width="1728" alt="Screenshot 2024-06-06 at 11 12 01" src="https://github.com/StampyAI/stampy-ui/assets/98026499/78fb9804-73ca-41e5-a474-8deb3fe3310b">


So it seems to be on the third or fourth prompt. But note that this doesn't happen with human-written responses, only bot responses.

melissasamworth:
Yep same

LeMurphant:
Issue cropped up with release candidate when I tested it

PR #735, opened 2024-06-18
Stampy redesign

mruwnik:
New chatbot UI deployment.

Testable version at https://redesign.aprillion.workers.dev/chat/


LeMurphant:
I can't vouch for the code, but the functionality looks good enough. 

The two issues that bug me most are
https://github.com/StampyAI/stampy-ui/issues/736
https://github.com/StampyAI/stampy-ui/issues/643
Still, I think it's ok for a first release.

app/root.css
```diff
@@ -192,145 +207,212 @@ h2 {
 }
 
 .padding-bottom-4 {
-  padding-bottom: var(--spacing-4);
+  padding-bottom: var(--spacing-4) !important;
```
Aprillion, line 210: I don't like overusing important ... but if no one has time to remove the wrong padding from selectors with higher priority that overwrite these classes, it's not a blocker 😭

app/root.css
```diff
@@ -340,7 +422,7 @@ a {
 
 a:hover {
   text-decoration: none;
-  color: var(--colors-teal-500, #1d9089);
+  /* color: var(--colors-teal-500, #1d9089); */
```
Aprillion, line 425: ?

app/root.tsx
```diff
@@ -133,7 +133,7 @@ function Head() {
   return (
     <head>
       <meta charSet="utf-8" />
-      <meta name="viewport" content="width=device-width,initial-scale=1" />
+      <meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1" />
```
Aprillion, line 136: huh, why do we need to disable zoom-out on mobile? there is no issue linked in the commit...

app/routes/questions.$questionId.$.tsx
```diff
@@ -132,8 +140,8 @@ export default function RenderArticle() {
             />
           }
         >
-          <Await resolve={Promise.all([question, tags])}>
-            {([resolvedQuestion, resolvedTags]) => {
+          <Await resolve={question}>
```
Aprillion, line 143: 👍

app/routes/questions.actions.tsx
```diff
@@ -149,20 +163,20 @@ export const Action = ({
     const response = await fetch('/questions/actions', {method: 'POST', body: searchParams})
 
     if (response.ok !== true) {
-      setActionTaken(!actionTaken)
+      // don't ask
+      setActionTaken(actionTaken)
```
Aprillion, line 167: ?!?

app/routes/questions.editors.tsx
Aprillion, line 1: why not deleting the file?

app/routes/questions.old.$questionId.tsx
Aprillion, line 1: another non-deleted file?

public/favicon-min-512.png
Aprillion, line 1: we deleted the code to switch between 2 favicons => this file shouldn't be re-added

public/favicon-min.ico
Aprillion, line 1: and this file is not needed either

Aprillion (dismissed):
✅ few comments, but I didn't notice anything that should block the release

app/hooks/useChat.ts
```diff
@@ -0,0 +1,293 @@
+export const CHATBOT_URL = 'https://chat.stampy.ai:8443/chat'
```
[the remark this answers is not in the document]
mruwnik: it should. But I gave up after trying various things and nothing working

app/routes/questions.actions.tsx
mruwnik: @zarSou9 explain yourself! For the allseeing gaze of Peter has fallen upon you!

Aprillion: I might have seen that problem [once or twice](https://stackoverflow.com/a/58877875/1176601)... I think the code is fine for that purpose (using `(v) => !v)` would be wrong if the handler is called multiple times), but the comment could improve:
```suggestion
      // if response is not ok, revert to original state as it was before this handler started
      setActionTaken(actionTaken)
```

bryceerobertson:
I think #643 is sufficiently problematic to be launch-blocking, since it seems to consistently happen at around the 3rd or 4th prompt. Other than that looks great.

bryceerobertson (approved):

zarSou9 (approved):
Looking great! Especially since #643 is fixed. 🙏 Nemo

....so who's gonna press it?

PR #737, opened 2024-06-19
Fix reference order

mruwnik:
The reference popups where showing incorrect contetn


app/components/Feedback/index.tsx
```diff
@@ -59,7 +59,7 @@ const Feedback = ({
       <CompositeButton secondary>
         <Action
           pageid={pageid}
-          showText={!!labels}
+          showtext={(!!labels).toString()}
```
Aprillion, line 62: I get `showText` vs `showtext` might have been a problem,, but what's with the `.toString()` part?

Aprillion (approved):

app/components/Feedback/index.tsx
mruwnik: `showtext` expects a string, so `false` was causing errors that `"false"` fixes...

Aprillion: :suspect:

Issue #717, opened 2024-06-11
Article suggestions refresh when user clicks settings


zarSou9:
https://github.com/StampyAI/stampy-ui/assets/148247390/583b5432-7add-41fd-bc90-c597d7180777

@mruwnik This still seems to be happening on occasion. e.g. "How does the Stampy chatbot work?" consistently has this bug

zarSou9:
It also happens on screen resize apparently:

https://github.com/StampyAI/stampy-ui/assets/148247390/dae08efa-7412-48e7-9f60-2cbe8379e2d3

This seems like a react thing which automatically updates the list which sounds very yucky

zarSou9:
Hmm, this is still an issue. The article for "How does the Stampy chatbot work?" is a good (or i guess bad) example. @mruwnik I can have a crack at it if you would like?

mruwnik:
give it a go

PR #740, opened 2024-06-20
fix: article suggestions refresh when user clicks settings #717

zarSou9:


app/components/Chatbot/index.tsx
```diff
@@ -145,13 +145,20 @@ type FollowupsProps = {
...
+    const initialItems =
+      (followups?.length || 0) >= 3
+        ? followups
+        : [
+            ...(followups || []),
+            ...randomQuestions().map(({title, pageid}) => ({text: title, pageid})),
+          ].slice(0, 3)
+
+    setItems(initialItems)
+    // eslint-disable-next-line
```
Aprillion, line 160: I would prefer if `randomQuestions` was a stable reference to a useCallback-ed function that wouldn't mess with useEffect deps,, but not a blocker if that's not a trivial fix (I haven't checked myself at the moment)

Aprillion (approved):

app/components/Chatbot/index.tsx
zarSou9: I agree, but yeah it might be difficult because `randomQuestions` comes from a hook -> `useOnSiteQuestions`

PR #747, opened 2024-07-16
Sundry fixes

mruwnik:
e.g. 
```
requests.get(
    'https://aistampy.info/questions/allQuestions',
    auth=HTTPBasicAuth('stampy', 'stampy'),
    params={'dataType': 'singleFileJson', 'questions': 'inProgress'}
).json()
```


app/components/search.tsx
```diff
@@ -31,6 +33,17 @@ export default function Search({limitFromUrl, className}: Props) {
     }
   }
 
+  useEffect(() => {
+    const query = searchParams.get('q')
+    if (loadedQuestions && query) {
+      searchFn(query)
+      setShowResults(true)
+    }
+    // Properly adding all dependancies would require transforming `searchFn` into a callback
```
Aprillion, line 42: from phone, but something like:
```ts
const searchFnRef = useRef(searchFn)
searchFnRef.current = searchFn
useEffect(
...searchFnRef.current(),
...
```

wrangler.toml.template
```diff
@@ -25,4 +25,4 @@ MATOMO_DOMAIN = "{MATOMO_DOMAIN}"
 DISCORD_LOGGING_CHANNEL_ID = "{DISCORD_LOGGING_CHANNEL_ID}"
 DISCORD_LOGGING_TOKEN = "{DISCORD_LOGGING_TOKEN}"
 EDITOR_USERNAME = "{EDITOR_USERNAME}"
-EDITOR_PASSWORD = "{EDITOR_PASSWORD}"
+EDITOR_PASSWORD = "{EDITOR_PASSWORD}"
```
Aprillion, line 28: ?

Aprillion (approved):

LeMurphant (approved):
I don't know about the code, but the format to request the pages looks good.

wrangler.toml.template
mruwnik: trailing whitespace

app/components/search.tsx
mruwnik: dunno if that's better... :D

wrangler.toml.template
Aprillion: yes, why removing it?

app/components/search.tsx
Aprillion: disabling the eslint rule is unsafe for future changes which can more easily [introduce bugs](https://react.dev/learn/removing-effect-dependencies#to-change-the-dependencies-change-the-code).. mutable refs might not look good, but you explicitly say "this is not a dependency which triggers a synchronization, I just want to use whatever value is has currently" without implying anything about any other variable that may or may not be a dependency

PR #760, opened 2024-08-29
Update frontpage suggestions.

LeMurphant:
I still think ideally the title should be fetched from the page directlly and not specified here

### DIFF
--- a/app/components/ContentBox/index.tsx
+++ b/app/components/ContentBox/index.tsx
@@ -72,13 +72,13 @@ export const ContentBoxSecond = () => {
       <ListTable
         sameTab
         elements={[
-          {title: 'Why can’t we just use Asimov’s Three Laws of Robotics?', pageid: '6224'},
           {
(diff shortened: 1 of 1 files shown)


mruwnik:
agreed :/

Also, you'll have to run `npm run lint:fix`

LeMurphant:
@mruwnik I ran the command and got an error in a file I did not modify. The error also shows up on the master branch.

`stampy-ui/node_modules/typescript/lib/tsc.js:93
  for (let i = startIndex ?? 0; i < array.length; i++) {
                           ^

SyntaxError: Unexpected token '?'
`

mruwnik:
huh. This is a typescript error, which is concerning. Have you updated your node modules recently? Does removing them and reinstalling work (`rm -R node_modules && npm install`)?

LeMurphant:
Ubuntu jammy had installed npm v8.5, I manually updated to v22.8.0 and it works.

mruwnik (approved):

PR #780, opened 2024-10-15
First draft for Knowledge

LeMurphant:
It's not perfect, but everything that seemed off got a TODO tag. @melissasamworth I'll need your help to fix them.

I'm pretty sure we still have spacing issues, but we can solve them once we've talked about the outline.


zarSou9:
Nice, yeah so @mruwnik updated how we are organizing the pages (for good reason), and added more components, but it means that this will need to change how they are written (should make it easier). Will let you know when it is time to move to other pages and you can update this. Sorry for the confusion

PR #786, opened 2024-10-17
careers page, various minor fixes #784

### LINKED ISSUE #784: careers page, various minor fixes
Myles will

- Add the missing sections
- Image component
  - Interested in pursuing this career path?
  - Testimonials
  - (don't worry about the articles section)
- Change the text style for this note
  - Image
- Put a hover state on the cards at the top
- Style the link blocks so that they're styling and hover state also match the cards
- Image
- Style the tag so that it uses the css class "xs"

zarSou9:


app/components/Testimonial/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  title,
+  description,
+  src,
+  layout = 'squeezed',
+}) => {
+  const isMobile = useIsMobile()
+  const finalLayout = isMobile ? 'squeezed' : layout
+  className = `testimonial ${finalLayout} ${className || ''}`
+
+  return finalLayout === 'squeezed' ? (
```
mruwnik, line 24: can't this be done with css + a media query?

app/components/HowCanIHelp/Career.tsx
```diff
@@ -0,0 +1,650 @@
+import Card from '../Card'
```
mruwnik, line 1: you're going to have to move this to `app/routes/how-can-i-help.career.tsx`, which will be a bit of a problem - sorry about that

app/components/Testimonial/index.tsx
zarSou9: Yeah, that was what I thought at first, but the layouts are just different. Like the puzzle doesn't fit. I can't articulate why, just look at the different versions and you can see what I mean.

app/components/HowCanIHelp/Career.tsx
zarSou9: All good, the change was needed. I've fixed it all now

app/components/Button/index.tsx
```diff
@@ -48,7 +48,7 @@ const Button = ({
   const LinkComponent = !action.startsWith('http')
     ? Link
     : ({to, children, ...props}: {[k: string]: any}) => (
-        <a href={to} {...props}>
+        <a href={to} rel="noreferrer" target="_blank" {...props}>
```
zarSou9, line 51: @bryceerobertson I switched it here so that by default links to other pages open in a new tab. @mruwnik is that ok?

mruwnik: not according to me. That's what the control button is for... But there was a whole discussion about this and I think the external links faction won :/ @melissasamworth is that correct?

app/components/HowCanIHelp/HelpItem.tsx
```diff
@@ -5,16 +5,21 @@ type HelpItemProps = {
...
       <div>
-        {title && <p className="default-bold padding-bottom-16">{title}</p>}
+        {title && <p className={`padding-bottom-16 ${titleFont}`}>{title}</p>}
         <p className="grey default padding-bottom-16"> {children} </p>
       </div>
       <div>
-        {links?.map((link) => <LinkCard key={link.title} action={link.action || '#'} {...link} />)}
+        {links?.map((link) => (
+          <div key={link.title} className="padding-bottom-16">
+            <LinkCard action={link.action || '#'} {...link} />
```
mruwnik, line 20: why not apply the class directly to the `LinkCard`?

app/components/Testimonial/index.tsx
mruwnik: that's unfortunate, but happens. @melissasamworth it's all your fault!!

mruwnik (approved):
I'd prefer the LinkCard to accept a `className` param, otherwise this looks good (other than the css, of course - I didn't check that)

app/components/HowCanIHelp/HelpItem.tsx
zarSou9: Because we are spacing with padding, and the padding would go inside the button

PR #790, opened 2024-10-18
Articles carousel

mruwnik:


zarSou9:
The width of the items shouldn't be fixed, they should fill the space. Just have `shown` be determined by the width of `componentRef` with decided intervals

mruwnik:
the problem with this is when there are 4 shown per partition, and there are 6 in total but you get to the last "page". I assume that there should only be 2 that take up 50% of the widget (in total), rather than streching to fill the space

app/components/HowCanIHelp/category-carousel.css
```diff
@@ -0,0 +1,43 @@
...
+  gap: var(--spacing-32);
+}
+
+.carousel-item {
+  flex: 1;
+  flex-direction: column;
+  align-items: start;
+  justify-content: start;
+  width: 100%;
+  max-width: 300px;
```
zarSou9, line 39: Ahh I see. You can just set this to a larger value, say 380px

zarSou9 (approved):

PR #809, opened 2024-10-25
images for video thumbnail

melissasamworth:


mruwnik (commented):
would it be possible to just merge them? Or is that a bad idea?

LeMurphant:
@mruwnik Merge this image with what?

bryceerobertson:
I think he means put the images together into one

melissasamworth:
@mruwnik I stubbornly don't want to because one is SVG the other is bitmap

mruwnik (approved):

PR #811, opened 2024-10-26
Add image key to LinkItems

mruwnik:
Still needs css:
image


app/routes/how-can-i-help.knowledge.tsx
```diff
@@ -28,6 +30,18 @@ const NewToAISafety = () => (
           {
             title: 'Intro to AI safety video playlist',
             action: 'https://aisafety.info/questions/9OGZ/',
+            image: (
+              <div style={{position: 'relative'}}>
```
mruwnik, line 34: this is very ugly and not the Right Way to do this kind of thing, but if it's a one off kind of thing it should be fine?

app/components/DropDown/dropdown.css
```diff
@@ -3,6 +3,10 @@
   border-bottom: 1px solid var(--colors-teal-200);
 }
 
+.drop-down:not(:last-of-type) {
```
mruwnik, line 6: this fixes the double line

bryceerobertson (approved):

PR #827, opened 2024-11-01
smooth scroll only on career

zarSou9:


mruwnik (approved):
I don't like how this is mucking around with the history. But I don't have a better idea :/

zarSou9:
> I don't like how this is mucking around with the history. But I don't have a better idea :/

yeah.. :/

PR #833, opened 2024-11-06
Remove [See more...] from backend

LeMurphant:
The feature has been broken for a while, the text was simply ignored by the parser. I have removed what I believe is all traces of it in the backend.

I have also made sure there are no articles (live or not) that contain the text **[See more** so we should not see any instances of this text showing up on the site. It has also been removed from all of our documentation.

There might be some legacy code to remove from GDocsRelatedThings, but if this is the case, I could not find it.

### DIFF
--- a/app/components/Article/Contents.tsx
+++ b/app/components/Article/Contents.tsx
@@ -180,11 +180,6 @@ const Contents = ({
       const target = e.target as HTMLElement
       target.classList.toggle('visible')
     }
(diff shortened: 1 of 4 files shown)


Aprillion (approved):
the changed code looks good to me (I haven't searched for leftovers of related code though, but if it works locally it should be fine when deployed ...  though please consider clearing cache after merging)

mruwnik (approved):

Issue #520, opened 2024-03-07
/categories: "page" instead of "pages" when there's just one article tagged


bryceerobertson:
![Image](https://github.com/StampyAI/stampy-ui/assets/98026499/257a6d82-9329-4ed0-a687-42c2bfd5f21a)

LeMurphant:
Solved by https://github.com/StampyAI/stampy-ui/pull/834

PR #834, opened 2024-11-06
Fix issue with results when there is only one page

LeMurphant:
Solves https://github.com/StampyAI/stampy-ui/issues/520

### DIFF
--- a/app/routes/categories.$.tsx
+++ b/app/routes/categories.$.tsx
@@ -54,6 +54,8 @@ export default function Tags() {
                   <div className="padding-bottom-24">
                     {selectedTag.questions.length === 0
                       ? 'No pages found'
+                      : selectedTag.questions.length === 1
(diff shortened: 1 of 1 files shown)


app/routes/categories.$.tsx
```diff
@@ -54,7 +54,9 @@ export default function Tags() {
                   <div className="padding-bottom-24">
                     {selectedTag.questions.length === 0
                       ? 'No pages found'
-                      : `${selectedTag.questions.length} pages tagged "${selectedTag.name}"`}
+                      : selectedTag.questions.length === 1
+                        ? `1 page tagged "${selectedTag.name}"`
+                        : `${selectedTag.questions.length} pages tagged "${selectedTag.name}"`}
```
mruwnik, line 59: this kind of nested tenary operations is cumbersome. Could you extract it out of the `return` into a proper `if` or something?

mruwnik (approved):

Issue #478, opened 2024-03-06
Footer: Line spacing between links too narrow


bryceerobertson:
Site:

![Image](https://github.com/StampyAI/stampy-ui/assets/98026499/e9c226ba-7bc0-417d-abbc-bd1fcfe2edcc)

Figma:

![Image](https://github.com/StampyAI/stampy-ui/assets/98026499/1f2d1ac6-0328-4a51-adfc-59d21f8875d1)

PR #835, opened 2024-11-06
Increase footer spacing

LeMurphant:
Ref https://github.com/StampyAI/stampy-ui/issues/478

@melissasamworth I tried using `<div className="margin-bottom-8">` instead but it did not work, for reasons I did not understand. I don't expect my fix to break anything else since *Link* is defined locally.


LeMurphant:
Melissa already fixed it
https://discord.com/channels/677546901339504640/1193299418259140690/1303884998067224648

Issue #674, opened 2024-06-01
Suggestion: Gradually replace AI generated images


ITR13:
Currently the site uses AI generated images in it's definition popups and "Advanced Sections" buttons, but doesn't have any page detailing if these are ethically sourced or what their training data was:
![image](https://github.com/StampyAI/stampy-ui/assets/4341642/f87748ba-10a3-453a-bdd0-b36570539854)
![image](https://github.com/StampyAI/stampy-ui/assets/4341642/6dd90161-dde1-4e08-ad60-95db3dfd5791)
![image](https://github.com/StampyAI/stampy-ui/assets/4341642/36d59311-fb81-432f-adc8-ca5b9dd15ddb)

The ethics behind generative AI is still under heavy debate, and a lot of people might view the site in a negative light if they spot AI generated images. People who are against generative AI are also likely to want safer AI, so keeping these images can be detrimental to the cause.

My suggestion is to gradually replace the AI generated images with ones contributed by artists. Preferably also doing steps to ensure it's easy to see what images still haven't been replaced, and making it easy for artists to contribute an image since not everyone knows how to use GitHub.

This of course excludes images like the ones on "[What is interpretability and what approaches are there?](https://aisafety.info/questions/8241/)" that are used as examples.

LeMurphant:
This is an interesting perspective. My (relatively uninformed) view was that the use of these images was debated if it was either 1) Emulating the style of a specific artist or 2) Replacing the work of an existing paid artist.

Given that it seems unlikely to me that we would pay artists to produce images for e.g. the glossary, the ethical constraints do not seem to apply as much to us. Still, if the perception is "AI-generated images -> bad", then it might be worth taking into consideration and switching to a paid bank of images.

At the moment, images for the glossary and icons for the sections are produced by prompting DALL-E, which I don't expect to have the best standards in terms of fair use of artists's work. If there is a platform that generates images that are more ethically sourced, I suggest we move to it even if the quality is lower. Unfortunately, if the people visiting the site cannot see the difference, this might not help the perception.

ITR13:
> This is an interesting perspective. My (relatively uninformed) view was that the use of these images was debated if it was either 1) Emulating the style of a specific artist or 2) Replacing the work of an existing paid artist.

[...]

Issue #553, opened 2024-03-15
Image compression in advanced section


LeMurphant:
See https://github.com/StampyAI/stampy-ui/issues/547#issuecomment-1995979234

Splitting to have different bugs from "inability to upload images"

Not launch-blocking

LeMurphant:
@melissasamworth there was some discussion that we might want to have icons instead of AI generated images, what do you think?

mruwnik:
ping @melissasamworth

bryceerobertson:
> @melissasamworth there was some discussion that we might want to have icons instead of AI generated images, what do you think?

I'm 100% for this

melissasamworth:
@bryceerobertson sorry, I suck at checking github notifications. 

YES. Let's please get rid of those terrible ai-gen images. I've just been lazy.

PR #837, opened 2024-11-06
Update advanced image ratio to 72x72

LeMurphant:
To address https://github.com/StampyAI/stampy-ui/issues/553

This is just a short-term fix, I think we might benefit from using icons instead of AI-generated images (see https://github.com/StampyAI/stampy-ui/issues/674), I might experiment with that.


LeMurphant:
Unfortunately it looks like I can't experiment with  the content of the images, since they are fetched directly from Coda, and changing them would update the images on the live website.

bryceerobertson:
I personally think AI-generated images there don't fit at all and look pretty ugly. Would love to see icons there like what's in the Figma

image

PR #852, opened 2024-11-18
Updating link

bryceerobertson:


bryceerobertson:
@mruwnik as things currently are, I'm going to have to update this link in this way every month 😬 Do you happen to have any easier way of updating it?

LeMurphant:
@bryceerobertson are you sure that this link does not automatically update to point to the next event? We could check if that is the case next month.

bryceerobertson:
@LeMurphant unfortunately it doesn't, seems to be a GCal bug

PR #853, opened 2024-11-18
Experimenting with blockquotes

LeMurphant:
Attempts at understanding why [blockquotes are not shown on site](https://github.com/StampyAI/GDocsRelatedThings/issues/64) even when they are tagged in markdown.

Created a page on http://localhost:8787/markdown to test markdown, but even when that works it looks like the CSS is not being loaded correctly on the actual pages, e.g. http://localhost:8787/questions/7602/Is-large-scale-automated-AI-persuasion-and-propaganda-a-serious-concern

In the end it does not work reliably, and I was not able to figure out why the CSS was not loading, but I thought the process might be relevant for the eventual solution.

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -19,6 +19,7 @@ import {CachedObjectsProvider} from '~/hooks/useCachedObjects'
 import {useTheme} from '~/hooks/theme'
 import {loadQuestionDetail} from '~/server-utils/stampy'
 import GlobalBanners from './components/GlobalBanners'
+import markdownStyles from '~/styles/markdown.css'
 
 /*
  * Transform the given text into a meta header format.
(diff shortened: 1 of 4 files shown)


LeMurphant:
Example of the markdown page 
markdown

LeMurphant:
I'm having inconsistent results, e.g. sometimes this page shows the blockquotes, sometimes it does not
http://localhost:8787/questions/7602/Is-large-scale-automated-AI-persuasion-and-propaganda-a-serious-concern
blockquote

Aprillion:
Re-implementing in a simplified way in https://github.com/StampyAI/stampy-ui/pull/872  ... no idea about "sometimes it does not" since I haven't observed it yet 🤷

LeMurphant:
Superceded by https://github.com/StampyAI/stampy-ui/pull/872

PR #855, opened 2024-11-24
improve error handling for not found ID

Aprillion:
* send a defered error message from loader to client (no point creating `new Response()` inside the deferred Promise on the client)
* https://stampy-ui2.aprillion.workers.dev/questions/bla


LeMurphant (approved):
Looks good, works as expected.

PR #856, opened 2024-11-24
support `/questions/1?list=1,2,3` manual list of ids

Aprillion:
* support manual list of ids like https://stampy-ui2.aprillion.workers.dev/questions/NM25/?list=NM25,9FQK,NM1S,8PYV
* discussed in https://discord.com/channels/677546901339504640/1125883086849188041/1309262887033901084
* on top of PR #855

### DIFF
--- a/app/components/ArticlesNav/articlenav.css
+++ b/app/components/ArticlesNav/articlenav.css
@@ -33,7 +33,7 @@
   flex: 1;
 }
 .mark-visited .articles-title:not(.selected) a:visited {
-  color: darkgray;
+  color: var(--colors-cool-grey-600);
 }
 
 .article-selector .dropdown-icon,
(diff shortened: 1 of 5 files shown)


LeMurphant:
Merged *not-found* into master, so changed this PR to merge into master. LMK if this was not the right way to proceed.

app/components/ArticlesNav/articlenav.css
LeMurphant, line 1: The color *darkgray* does not exist elsewhere on the site (neither does a9a9a9), so it should use the predefined color. I'm not sure what the color of visited links is, or where it is defined, but it seems like both this page and the rest of the site should use a similar option like *mark-visited*.

LeMurphant (commented):
I think having the *next* button point to the next article in the custom list (and not the regular hierarchy) might make sense, but it's not necessary in an MVP.

I'll let @mruwnik review the structure and Rob review the functionality before I merge

LeMurphant (approved):
Looks good to me with the updates. I'll leave 24 hours for @robertskmiles and @mruwnik to add comments if they want, and then I will merge.

mruwnik:
seems fine. I don't like this kind of hackery, but don't have a better solution :/

Issue #868, opened 2024-12-16
ToC is broken


LeMurphant:
See https://github.com/StampyAI/stampy-ui/commit/f3dbae4f1299b69afc28b9552de7c0c990df490a#commitcomment-150335993

PR #870, opened 2024-12-23
fix ToC bug

mruwnik:
Should fix [this bug](https://github.com/StampyAI/stampy-ui/issues/868)


LeMurphant (approved):
I don't have a good understanding of what the code does, but it fixes the issue.

Issue #863, opened 2024-12-05
Make internal links clearer


Algon-33:
Based on some [feedback](https://ruddy-jumpsuit-a19.notion.site/Feedback-Claire-D-15205b2ef69580fa8ec2c31e0147fd31) from Murphant's friend Claire, it might be a good idea to differentiate links to stampy articles vs external links. One way to do this is to mimic Gwern and place a little stampy symbol after links to our articles. See the image below for an example of Gwern doing this. 
![image](https://github.com/user-attachments/assets/bc47ab51-d2bc-4fc0-aad5-d830524dddcd)

@melissasamworth

LeMurphant:
Huh, it looks like the *external link* functionality used to exist on the old site https://github.com/StampyAI/stampy-ui/issues/863

I'd rather we recreate the old functionality (external box on external links) rather than tag internal links. @Algon-33 said that he would like to have something for internal links as well. I will recreate the external link box, and we can discuss internal links after.

LeMurphant:
It looks like removing the icon for external links was on purpose, see https://github.com/StampyAI/stampy-ui/pull/346

@Aprillion I don't understand the context for removing this functionality. Is it something that depended on the "cards" functionality of the old UI? Does it make sense to revert this commit? Would it break aisafety.com?

Aprillion:
@LeMurphant at the time, internal links opened the question "in-place" - the site was a list of questions that could be expanded or collapsed and a link would add an expanded question to your list ... while external links opened on new tab and the icon was there to distinguish between the 2 behaviours

by removing the in-place feature and opening all links on new tab, we didn't need to distinguish between 2 behaviours, all links behaved the same after #346

feel free to copy&paste reusable parts of the deleted code into appropriate places in new codebase + test how it works, but please don't just revert the git commit (I would expect we now have different file structure so it wouldn't work.. on the other hand you can try and see and then just not push if it won't work)

LeMurphant:
Thanks for the context @Aprillion. My plan was to use the icon you were using previously to indicate not only that this would open a new page (all links do this) but that this links to a website that is not aisafety.info. Do you think it would be confusing to have this icon used in this way? Would another icon be better to denote "this is not on our website"?

Aprillion:
Should be fine to reuse the currently-unused icon for new purpose, no opinion if it's the best one to choose from - maybe @melissasamworth has an opinion.

PR #871, opened 2025-01-02
Add icon to external links

LeMurphant:
fixes https://github.com/StampyAI/stampy-ui/issues/863

There were a few modifications compared to what was done in https://github.com/StampyAI/stampy-ui/pull/346, which Claude says give better compatibility, but @Aprillion LMK if it's wrong

external_icon

### DIFF
--- a/app/root.css
+++ b/app/root.css
@@ -640,18 +640,27 @@ a:not(:hover, :focus-visible) {
   text-decoration: none;
 }
 
+.contents a[target='_blank'][href^='http']:not(.icon-link, .transparent-link) {
+  margin-right: 12px;
+}
+
 .contents a[target='_blank'][href^='http']:not(.icon-link, .transparent-link):after {
-  content: ' ';
-  display: inline-block;
+  content: '';
+  position: absolute;
   width: 10px;
   height: 10px;
   margin-left: 1px;
-  margin-bottom: 6px;
+  margin-top: 8px;
   -webkit-mask: url('/assets/Icon_External_Link.svg') no-repeat center center;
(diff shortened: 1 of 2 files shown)


LeMurphant:
Unfortunately, I have noticed that in rare cases, the icon can appear after the line-wrap. I was not able to fix this.
line_break

Aprillion:
> Unfortunately, I have noticed that in rare cases, the icon can appear after the line-wrap. I was not able to fix this.

I pushed a commit with `position:absolute` fix/workaround, feel free to test on https://stampy-ui2.aprillion.workers.dev/questions/6953/Do-people-seriously-worry-about-existential-risk-from-AI and revert if not good for all edge cases.

I also fixed line height of lines with footnotes using `<sub>` to stay the same as lines without footnotes:

external icon not wrapping and line height

Aprillion (approved):

LeMurphant:
@Aprillion your fix works on Firefox (it did not work before), but on Chrome I still get the same result as before.

Aprillion:
hmm, previous version looked good in Safari too ... maybe Chrome optimized performance of some more reasonable CSS and not for absolutely positioned pseudoelements inside inline line-wrappable parents or something like that 🙈 

now using a different workaround, to bake in the background directly to the element without `:after` - but now we can only choose 1 color, I chose the visited link color because it looked better for both unvisited and visited links (while the unvisited color looked strange for visited links) ... feel free to choose another fixed color if this workaround wins

LeMurphant:
@Aprillion this looks good to me, do you mind re-approving the PR so I can merge?

Aprillion (approved):

Issue #876, opened 2025-02-04
Feature request: Carousel


Algon-33:
@zarSou9 
Here's a feature request for the new intro section the content team is working on. Basically, we want to be able to make a carousel so that we can show multiple images, gifs and embedded videos in the same bit of the page. Here's an example which comes from [this page](https://jvnhair.com/products/revive-rapid-repair-bond-gloss).
![Image](https://github.com/user-attachments/assets/1b731bb8-406c-4127-8add-85ddf3c8b632)
That is, we want some interactive element on article pages that lets users click through several possible images/gifs. I'd prefer there to be no automatic movement through the options in the carousel. The ability to put gifs, embedded videos, and images in the carousel would be great. Ideally, this would be integrated with the Google Docs that we write our articles in. 

This page describes best practices for carousels: https://blog.hubspot.com/website/website-carousel-best-practices

LeMurphant:
It's worth noting that this must apply to pages that have been written in Google Docs. This means we need a pipeline from Google Docs -> Markdown -> Shown on the website.

LeMurphant:
Closed by https://github.com/StampyAI/stampy-ui/pull/879

Issue #766, opened 2024-09-17
No way to center image captions


LeMurphant:
See for instance 
![francis](https://github.com/user-attachments/assets/7fa2e2b2-6400-4eb8-b7b1-548746b95cae)
and
![francis_doc](https://github.com/user-attachments/assets/07de08b3-0dcd-4644-a70e-054b607a81db)
from https://aisafety.info/questions/NM1M/Isn't-the-real-concern-deepfakes

It's possible that centering is not the best way to use these captions, I'm open to some better way.

mruwnik:
this would probably require the parser to add custom html around these images... Should all images be centered, or only ones specifically centered in gdocs? Having all images be centered is a lot simpler, but is a a Design Choice

LeMurphant:
I can't think of a situation where we would want a non-centered image, so go for centering them all.

We would want to center the caption as well. More generally, it seems like a good idea to test for centered text in Gdocs in the same way [we test for blockquotes](https://github.com/StampyAI/GDocsRelatedThings/pull/73). I could write the code to parse it, but is it ok to add the HTML directly with something like this or will it get stripped?
```
<div style="text-align: center">
This text will be centered
</div>
```

PR #879, opened 2025-02-05
Media carousel from feature request #876

zarSou9:
This adds the ability to create media carousels in articles using [this format](https://docs.google.com/document/d/12oF6ehIHvf3ihORypTwYkE21dCFsz4yF6Ewwu6FDllk/edit?tab=t.0) from google docs. This (I'm pretty sure, but @nemo can you verify) translates in markdown to something like the following:
```
::: carousel
- [Cat Image Title](https://media.4-paws.org/d/2/5/f/d25ff020556e4b5eae747c55576f3b50886c0b90/cut%20cat%20serhio%2002-1813x1811-720x719.jpg)

- [Cat Gif Title](https://i.giphy.com/SKDiTL793TGiaiV9Kv.webp)

- [Cat Video Title](https://www.youtube.com/watch?v=cytJLvf-eVs)

:::
```

Which is then parsed, and replaced with a media carousel component which looks like this:

https://github.com/user-attachments/assets/8cdaa09b-5b96-4a8d-874d-ea0905ca2d87


LeMurphant:
There are a lot of extra spaces added, but markdown output looks ok on
https://docs.google.com/document/d/10g6U9SL0CBy__wCBTib7_WhB3S3aaFt7Fx1vVgCzg2I/edit?tab=t.0

`::: carousel


- [Cat Image Title](https://media.4-paws.org/d/2/5/f/d25ff020556e4b5eae747c55576f3b50886c0b90/cut%20cat%20serhio%2002-1813x1811-720x719.jpg)

- [Cat Gif Title](https://i.giphy.com/SKDiTL793TGiaiV9Kv.webp)

- [Cat Video Title](https://www.youtube.com/watch?v=cytJLvf-eVs)

:::
`

LeMurphant:
View on site works mostly as expected. I suggest making the left-right arrows a bit thicker.
caroussel


Interface on mobile is not ideal, but I think it's fine. We can improve it if we end up using it extensively.

@Algon-33 do you have the tools to test it in a dev branch?

LeMurphant:
I don't know how hard that is, but it would be nice to be able to have captions that appear under the images to clarify what they are.

zarSou9:
> I don't know how hard that is, but it would be nice to be able to have captions that appear under the images to clarify what they are.

Agreed, I've just added it

LeMurphant:
caroussel2

Looks good to me!

Algon-33 (changes requested):
As I said elsewhere: 
"By the way, are the captions in that GD just "Cat Image Title" etc?

Two minor issues:

    there's no counter saying how many images there are and which one you're on. Displaying that as something like "1/3" would be good.
    It would be nice if you could click left/right on the first/last image so the carousel can wrap-around."

LeMurphant:
(for Github) Fix https://github.com/StampyAI/stampy-ui/pull/879

@Algon-33 yes the site shows the captions that are found in the google docs, see the input near the bottom of this document
https://docs.google.com/document/d/10g6U9SL0CBy__wCBTib7_WhB3S3aaFt7Fx1vVgCzg2I/edit?tab=t.0

I agree with both recommendations

LeMurphant:
(for Github) Fix https://github.com/StampyAI/stampy-ui/issues/876

zarSou9:
lmk what you guys think:

https://github.com/user-attachments/assets/5aa41044-4cbb-49a5-afc7-c310a2c8f062

Algon-33 (dismissed):
Looks good. Murphant helped me setup the ui locally, and the carousel looks great. Thanks!

zarSou9:
Thank you! Also, sorry I made another commit so someone needs to approve it again to merge. And @mruwnik, @melissasamworth, or @Aprillion, feel free to review this for code quality.

app/components/Article/Contents.tsx
```diff
@@ -173,6 +178,15 @@ const Contents = ({
     const el = elementRef.current
     if (!el) return
 
+    // Replace carousel placeholders with actual carousel components
+    carousels?.forEach((carousel) => {
+      const placeholder = el.querySelector(`#${carousel.id}`)
+      if (placeholder) {
+        const root = createRoot(placeholder)
```
Aprillion, line 185: not sure about this, but no quick suggestions and shouldn't cause any problems if we don't plan too many carousels per page 😓 ... we don't plan multiple carousels per page, right?

app/components/MediaCarousel/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+import {Navigation} from '../CategoryCarousel'
+
+type MediaCarouselProps = {
+  items: MediaItem[]
+}
+
+const MediaCarousel = ({items}: MediaCarouselProps) => {
+  const [currentIndex, setCurrentIndex] = useState(0)
+
+  const renderMedia = (item: MediaItem) => {
```
Aprillion, line 13: (minor) looks like a pure function => should be possible to move outside of the component ... or maybe even convert to real component `<Media item={items[currentIndex]} />`

app/server-utils/parsing-utils.ts
```diff
@@ -85,3 +85,66 @@ export const allLinksOnNewTab = (html: string): string => {
...
 }
+
+// Helper function to convert YouTube URLs to embed URLs
+const resolveYoutubeEmbedUrl = (url: string): string => {
+  // Pattern to get youtube ID found from https://regex101.com/library/OY96XI
+  const pattern =
+    /(?:https?:)?(?:\/\/)?(?:[0-9A-Z-]+\.)?(?:youtu\.be\/|youtube(?:-nocookie)?\.com\S*?[^\w\s-])([\w-]{11})(?=[^\w-]|$)(?![?=&+%\w.-]*(?:['"][^<>]*>|<\/a>))[?=&+%\w.-]*/gim
+  const match = pattern.exec(url)
+
+  if (match && match[1]) return `https://www.youtube.com/embed/${match[1]}`
```
Aprillion, line 96: this was tested and works fine in local deployment, right? (I mean not just in regex101, but on a real answer page with youtube link)

Aprillion (approved):
if all works when tested, LGTM

app/server-utils/parsing-utils.ts
LeMurphant: Yeah it works fine

app/components/Article/Contents.tsx
LeMurphant: I had not really thought of that option, but I think it's fine if we restrict ourselves to 1 carousel per page. This does make https://github.com/StampyAI/stampy-ui/issues/766 a bit more relevant though because I expect we will sometimes want more than 1 non-contiguous image per page.

app/components/MediaCarousel/index.tsx
LeMurphant: @zarSou9 lmk if you want to fix it before I merge

app/components/Article/Contents.tsx
Aprillion: yeah, carousel should have multiple images to carouse around 🎠 and shouldn't be used for single images => I will look into centering the normal images 🤞

app/components/MediaCarousel/index.tsx
zarSou9: > (minor) looks like a pure function => should be possible to move outside of the component ... or maybe even convert to real component `<Media item={items[currentIndex]} />`

Yes good point, fixed

zarSou9: Ok @LeMurphant you can approve and merge now

Aprillion (approved):

PR #886, opened 2025-02-15
center images and captions

### LINKED ISSUE #766: No way to center image captions

Aprillion:
* fixes #766
* https://stampy-ui2.aprillion.workers.dev/questions/NM1M/

### DIFF
--- a/app/components/Article/Contents.tsx
+++ b/app/components/Article/Contents.tsx
@@ -187,6 +187,16 @@ const Contents = ({
       }
     })
 
+    // Process captions
+    const paragraphs = el.getElementsByTagName('p')
+    for (const p of paragraphs) {
+      const text = p.textContent || ''
+      const html = p.innerHTML || ''
+      if (text.startsWith('Caption:')) {
+        p.classList.add('image-caption')
+        p.innerHTML = html.replace(/Caption:\s*/, '')
+      }
+    }
     updateTextNodes(el, insertGlossary(pageid, glossary))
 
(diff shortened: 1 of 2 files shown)


LeMurphant:
This works as expected, but centering is often not pretty. I will keep the mechanism of "apply formatting to after image" and attempt to find a nicer way to format it before merging.

LeMurphant:
After some feedback from the team, we decided to use a `Caption: ` mechanism instead
caption

Aprillion:
https://stampy-ui2.aprillion.workers.dev/ is updated .. I have no idea which document uses the new caption to test, but updated code looks good to me

LeMurphant:
@Aprillion you can see the result on these pages
https://stampy-ui2.aprillion.workers.dev/questions/8241/What-is-interpretability-and-what-approaches-are-there
https://stampy-ui2.aprillion.workers.dev/questions/0
The former has an issue with the second image, but it seems unrelated to the PR, it has the same issue on the live site

Aprillion:
hmm, the image loading seems broken 😱 page refresh helps as a workaround ... but let's investigate more => https://github.com/StampyAI/stampy-ui/issues/887

LeMurphant (approved):
Now words as expected

Issue #659, opened 2024-05-28
Favicon needs transparent background


bryceerobertson:
![Image](https://github.com/StampyAI/stampy-ui/assets/98026499/8efb1112-0054-49e0-b0ad-06a23c3f15d0)

bryceerobertson:
Looks great

PR #888, opened 2025-03-02
Update favicon with transparency

LeMurphant:
Fixes https://github.com/StampyAI/stampy-ui/issues/659

light_dark

This increases the size a bit (from 5k to 16k) since it contains a higher resolution. It also loses the 24px resolution that was present before, but that does not seem to be an issue.

Another option would have been to use favicon-512.png, but that bumped size to 50k.

@Aprillion I noticed this comment you wrote
```
      {/* don't use color-scheme because supporting transparent iframes https://fvsch.com/transparent-iframes
          is more important than dark reader https://github.com/darkreader/darkreader/issues/1285#issuecomment-761893024
          <meta name="color-scheme" content="light dark" />
       */}
```
It looks like that won't interfere, but I wanted to double-check with you.


public/favicon.ico
Aprillion, line 1: aisafety.com no longer uses embedded / iframe version of aisafety.info, so there would be no issue with automatic dark mode ... but also the comment was from a time when we had dark mode version of the site 🖤  feel free to test with Dark Reader browser extension whether it looks better with/without that particular `<meta>` tag ... not related to favicon at all

Aprillion (approved):

Issue #656, opened 2024-05-28
chatbot input is one line only


mruwnik:

mruwnik:
@melissasamworth what should happen if a user inputs a very large string into the chatbox input? Should it go multiline, or start scrolling?

bryceerobertson:
I strongly think it should go multiline like ChatGPT 

![Image](https://github.com/StampyAI/stampy-ui/assets/98026499/4f3ecded-1363-4cd8-b603-8ce4bb1c2450)

PR #918, opened 2025-04-01
Make Chat input multiline

LeMurphant:
Fix https://github.com/StampyAI/stampy-ui/issues/656

single-line
three-lines
scrollbar

There are a bunch of design decisions that ideally would have been taken around the position of the "send" key, I did something that made sense to me but I'm open to changing it

I also notice that when the text area expands, it eats up the questions above it. I think it's minor enough that it's fine.

### DIFF
--- a/app/components/Chatbot/widgit.css
+++ b/app/components/Chatbot/widgit.css
@@ -69,8 +69,11 @@
 
 .widget-ask .send {
   position: absolute;
-  right: var(--spacing-8);
-  bottom: var(--spacing-8);
+  right: var(--spacing-16);
+  top: 32px;
+  transform: translateY(-50%);
+  z-index: 5; /* Ensure the button stays above the scrollbar */
+  transition: top 0.1s ease; /* Match textarea's transition */
 }
 
 .white-space {
--- a/app/components/Input/index.tsx
+++ b/app/components/Input/index.tsx
@@ -1,4 +1,5 @@
 import './input.css'
+import { useRef, useEffect, RefObject } from 'react'
 
 type InputProps = {
   className?: string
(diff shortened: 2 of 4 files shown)


bryceerobertson:
Lol I appreciate the lorem ipsum. I think the send button should go at the bottom instead of the top. That seems to be closer to the standard. Otherwise seems great.
image

LeMurphant:
@bryceerobertson how about now?

button-bottom

LeMurphant:
@Aprillion adding you as a reviewer for feedback on CSS styling

app/components/Chatbot/widgit.css
```diff
@@ -69,8 +69,10 @@
 
 .widget-ask .send {
   position: absolute;
-  right: var(--spacing-8);
-  bottom: var(--spacing-8);
+  right: var(--spacing-16);
+  bottom: 24px;
+  z-index: 5;
```
Aprillion, line 74: does z-index do anything for a component with position absolute? (see https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context)

app/components/Input/index.tsx
```diff
@@ -8,11 +9,56 @@ type InputProps = {
...
+  textareaRef?: RefObject<HTMLTextAreaElement>
 }
-const Input = ({className, ...props}: InputProps) => {
+const Input = ({className, multiline, textareaRef, ...props}: InputProps) => {
   const classes = ['input', className].filter((i) => i).join(' ')
+  const internalRef = useRef<HTMLTextAreaElement>(null)
 
-  return <input className={classes} {...props} />
+  // Use provided ref or internal ref
+  const textareaReference = textareaRef || internalRef
```
Aprillion, line 20: hmm, I don't understand what is purpose of this "mixin" ref 🤔

app/components/Input/index.tsx
```diff
@@ -8,11 +9,56 @@ type InputProps = {
...
+
+  // Auto-resize textarea based on content
```
Aprillion, line 22: I can see `react-auto-height` is still in package.json but not used in any component any more :(

does this effect work better than https://peter.hozak.info/react-auto-height/?path=/story/autoheight--textarea-controlled-via-render-props ?
* if this custom code is better, feel free to remove the library from dependencies
* if it works the same, I would prefer the library I wrote where I handled a bunch of edge cases... (though maybe not those we need to handle here, it was long time ago)

Aprillion (approved):
no blockers AFAICT

bryceerobertson:
LGTM

app/components/Input/index.tsx
LeMurphant: Honestly, all of this is outside of my wheelhouse. I massaged the code (with Claude's help) until it worked with react-auto-height (both in Chrome and Firefox), but feel free to suggest updates to the way I integrated it.

app/components/Input/index.tsx
```diff
@@ -8,10 +10,17 @@ type InputProps = {
...
+  multiline?: boolean
 }
-const Input = ({className, ...props}: InputProps) => {
+
+const Input = ({className, multiline, ...props}: InputProps) => {
   const classes = ['input', className].filter((i) => i).join(' ')
 
+  if (multiline) {
+    // Use createElement to pass direct element prop to AutoHeight
+    return createElement(AutoHeight, {element: 'textarea', className: classes, ...props})
```
Aprillion, line 21: whoa, what's this? 👀  I will look into this ~tomorrow, but this code looks.. unexpected

Aprillion: fixed in https://github.com/StampyAI/stampy-ui/pull/918/commits/ac614c0395494620e2e74dd08c3fbfb0ff8cf5ef

Screenshot 2025-04-10 at 9 46 10

Aprillion (approved):
(approving to unblock merge after @LeMurphant will revers-review my commit 😅)

LeMurphant:
@Aprillion I pinned the "send" button to the bottom as Bryce asked.

some-text

My version had a maximum textArea zone of 3x the initial size which was suggested by the LLM, but I think it's fine to have it expand indefinitely.

I notice that the spacing from the bottom is different in Chrome vs Firefox. I optimized it for Chrome, but if there's a nicer way, I'm open to changing it.
chrome
firefox

Aprillion:
yeah, the Chrome-Firefox difference was why I chose to pin the button from top as the simplest quick workaround

I'd be fine if someone else will approve the current version broken in Firefox (or investigate a version that pins the button from bottom which would work correctly in both browsers), but I'm too stubborn to approve a version broken in Firefox myself, sorry 🦊

LeMurphant:
@Aprillion it feels very un-principled, but the only option I got to work was simply having a different bottom padding for Firefox. The result looks OK, but I wish there was a nicer way.

Chrome is on the left, Firefox on the right
chrome_firefox

Aprillion (approved):
thank you!

Issue #923, opened 2025-04-07
Em-dashes sometimes displayed as hyphens


srkaas:
Em-dashes on the site are sometimes displayed as hyphens. It seems to vary unpredictably (sometimes it happens when loading the same page, sometimes not). Attached screenshot is from Chrome on Windows.

![Image](https://github.com/user-attachments/assets/96eaa9b3-59b1-40af-b420-f3fca78587e8)

PR #926, opened 2025-04-08
Update glossary to no longer replace text if there are no replacements necessary

LeMurphant:
fix https://github.com/StampyAI/stampy-ui/issues/923

It turns out that there was em-dash and quote replacement that was meant to normalize the glossary but applied even if there were not glossary hits. This led to seemingly random replacement of em-dashes to hyphen that would change upon refresh, which has annoyed @srkaas [for](https://discord.com/channels/677546901339504640/1125883308979535922/1300700976763899924) some [time](https://discord.com/channels/677546901339504640/1209956074984181830/1221869562035699774)

### DIFF
--- a/app/components/Article/Contents.tsx
+++ b/app/components/Article/Contents.tsx
@@ -86,7 +86,7 @@ const normalizeForComparison = (text: string): string => {
   return text
     .replace(/[‘’]/g, "'") // Replace left and right single quotes (U+2018, U+2019)
     .replace(/[“”]/g, '"') // Replace left and right double quotes (U+201C, U+201D)
-    .replace(/[–—]/g, '-') // Replace en-dash and em-dash with hyphen
+    .replace(/[‒–—]/g, '-') // Replace figure-dash, en-dash and em-dash with hyphen
 }
 
 const glossaryInjecter = (pageid: string, glossary: Glossary) => {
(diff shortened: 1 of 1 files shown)


app/components/Article/Contents.tsx
```diff
@@ -81,20 +81,37 @@ const updateTextNodes = (el: Node, textProcessor: (node: Node) => Node) => {
  *  - an on hover popup with a short explaination of the glossary item
  *  - use each glossary item only once
  */
+const normalizeForComparison = (text: string): string => {
+  // Standardize single quotes and replace em-dashes with hyphens
+  // Use Unicode escape sequences for better editor compatibility
+  return text.replace(/\u2019/g, "'").replace(/\u2014/g, '-')
```
Aprillion, line 87: I don't think there is such thing as a need for unicode escapes instead of real unicode characters for "editor compatibility" in 2025

but I suspect we should replace both left and right "fancy" quotation marks => `/[\u2018\u2019]/g` or `/[‘’]/g`, whatever actually works better in regex

also what about en-dash and figure-dash, those are not arriving from gdocs in need of replacement?

Aprillion (approved):

app/components/Article/Contents.tsx
LeMurphant: I added that because Cursor was complaining about a previous iteration having the wrong number of arguments, but it no longer seems to do so, and I will trust your judgement on that.

I updated to deal with fancy quotation marks, en-dash and figure-dash (I did not know that was a thing). In practice, I don't expect this to be an issue.

Aprillion (approved):

Issue #928, opened 2025-04-09
Glossary capitalizes terms in the main text


srkaas:
Terms in the main text that were uncapitalized in the Google Doc are shown capitalized on the site in some cases if they have a glossary entry.

For example, from https://aisafety.info/questions/NM3Q/Intro-to-AI-safety :

![Image](https://github.com/user-attachments/assets/88984910-f9b5-4471-b07f-c2501ed5883e)

PR #932, opened 2025-04-13
Ensure the original glossary text is kept

LeMurphant:
Fix https://github.com/StampyAI/stampy-ui/issues/928

### DIFF
--- a/app/components/Article/Contents.tsx
+++ b/app/components/Article/Contents.tsx
@@ -107,13 +107,9 @@ const glossaryInjecter = (pageid: string, glossary: Glossary) => {
         if (!seen.has(term) && normalizedHtml.search(match) >= 0) {
           seen.add(term)
 
-          // Preserve original text while marking as glossary term
-          const originalMatchRegex = new RegExp(`(^|[^\\w-])(${normalizedAlias})($|[^\\w-])`, 'i')
-
-          return currentHtml.replace(
-            originalMatchRegex,
-            (_, g1, g2, g3) => `${g1}<span class="glossary-entry">${g2}</span>${g3}`
-          )
+          // Use the canonical glossary term for consistent formatting
+          // Apply the replacement directly to the current HTML
+          return currentHtml.replace(match, `$1<span class="glossary-entry">${alias}</span>$3`)
         }
         return currentHtml
       }, html)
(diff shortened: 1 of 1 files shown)


app/components/Article/Contents.tsx
```diff
@@ -107,9 +107,13 @@ const glossaryInjecter = (pageid: string, glossary: Glossary) => {
         if (!seen.has(term) && normalizedHtml.search(match) >= 0) {
           seen.add(term)
 
-          // Use the canonical glossary term for consistent formatting
-          // Apply the replacement directly to the current HTML
-          return currentHtml.replace(match, `$1<span class="glossary-entry">${alias}</span>$3`)
```
Aprillion, line 112: would it be possible to make a prediciton (with or without AI assistants) whether this code diff looks reasonable or whether one more "are you sure" might have been appropriate? (hint: notice line 104)

Screenshot 2025-04-14 at 9 04 18

* after starting again from the original code, could you please delete the 2 comment lines again, but then only replace `{alias}` with a random integer between 1 and 3? 🙏  thank you!
* if we have some prettier/eslint config for not using template literals unnecessarily, please also format it and it might end up with single quotes instead of backticks:
`'$1<span class="glossary-entry">$2</span>$3'`

_(I'm on vacation this week and promised myself not to open code editor quite yet, but let me know if my proposal doesn't work and I will invetigate more later)_

Aprillion (changes requested):

app/components/Article/Contents.tsx
LeMurphant: Ok, not my finest moment :sweat_smile: 

I was not aware that single quotes were preferred to template literals when the latter are not necessary, I have updated the formatting. 

.prettierrc has ` "singleQuote": true` but it makes sense that it can't know when template literals are necessary.

Also, Claude dies not pick up on sarcasm very well yet :grin: 
```
● Now I need to replace {alias} with a random integer between 1 and 3 on line 104. Let me check the current code:

│ │ 104          const match = new RegExp(`(^|[^\\w-])(${normalizedAlias})($|[^\\w-])`, 'i')                                                                                                                      │ │
│ │ 104          const match = new RegExp(`(^|[^\\w-])(${Math.floor(Math.random() * 3) + 1})($|[^\\w-])`, 'i')  
```

Aprillion (approved):

PR #935, opened 2025-04-13
Fix glossary hard to escape

### LINKED ISSUE #929: Hard to escape glossary
On window widths around 1000px and less, it's quite hard to close the glossary pop-up that takes over the whole window on hover 😱 

Image

zarSou9:
This fixes #929

Alright this is the best I could do. The issue was that if the user starts out on a wider screen and then shortens it, the glossary popup gets added twice, with both desktop and mobile event handlers. I also fixed an issue where you would have to click multiple times to escape a modal on mobile.

I also notice a discrepancy between what the decided mobile with is in the root.css and the `useIsMobile` hook (is there a reason for this?), I updated the hook to be consistent with the css, because it was causing weird formatting for the glossary when the screen was in between the different values. However `useIsMobile` is used in many other places, so I don't know if this breaks anything.

### DIFF
--- a/app/hooks/isMobile.tsx
+++ b/app/hooks/isMobile.tsx
@@ -2,7 +2,7 @@ import {useState, useEffect} from 'react'
 
 const isWindow = typeof window !== 'undefined'
 export default function useIsMobile() {
-  const mobileWidth = 780
+  const mobileWidth = 1136
   const getWidth = () => (isWindow ? window.innerWidth : 1280) // assume desktop until proven otherwise (for SSR)
   const [windowWidth, setWindowWidth] = useState<number>(getWidth)
 
--- a/app/hooks/isMobile.tsx
+++ b/app/hooks/isMobile.tsx
@@ -2,7 +2,7 @@ import {useState, useEffect} from 'react'
 
 const isWindow = typeof window !== 'undefined'
 export default function useIsMobile() {
-  const mobileWidth = 1136
+  const mobileWidth = 780
   const getWidth = () => (isWindow ? window.innerWidth : 1280) // assume desktop until proven otherwise (for SSR)
   const [windowWidth, setWindowWidth] = useState<number>(getWidth)
 
--- a/app/hooks/isMobile.tsx
+++ b/app/hooks/isMobile.tsx
@@ -1,8 +1,7 @@
 import {useState, useEffect} from 'react'
 
 const isWindow = typeof window !== 'undefined'
-export default function useIsMobile() {
-  const mobileWidth = 780
+export default function useIsMobile(mobileWidth = 780) {
   const getWidth = () => (isWindow ? window.innerWidth : 1280) // assume desktop until proven otherwise (for SSR)
   const [windowWidth, setWindowWidth] = useState<number>(getWidth)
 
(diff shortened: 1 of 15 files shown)


app/hooks/isMobile.tsx
```diff
@@ -2,7 +2,7 @@ import {useState, useEffect} from 'react'
 
 const isWindow = typeof window !== 'undefined'
 export default function useIsMobile() {
-  const mobileWidth = 780
+  const mobileWidth = 1136
```
Aprillion, line 5: code looks fine 👍 

but this number is way too big IMHO, I would prefer to have all mobile CSS up to 780px, not on my laptop while trying to use over-half-screen window width

@melissasamworth, @bryceerobertson, @LeMurphant - are you sure the screenshot in #929 looks good to you - to have take-over-the-whole-page modal pop-up at that window size?  (either way, this PR would be a big improvement, I would just prefer to synchronize the other CSS with this "reasonable" number, not the other way round ... but I don't have background why we changed the CSS over time)

Aprillion (approved):

app/hooks/isMobile.tsx
LeMurphant: It looks too wide to me in that screenshot. But we have had issues with not having enough space to display the Glossary definitions, so I don't want to compress the text area to under half the size it uses in this screenshot.

bryceerobertson: Agreed that screenshot looks bad

zarSou9: Yeah it isn't great. Though when I tried switching all mobile breakpoints to only 780px, I got another (I think bigger) problem...

https://github.com/user-attachments/assets/38ce9ca8-f28e-4164-b9a3-85243da2de82

Though I think it should be consistent across the codebase (I.e. I'm seeing other occurrences of the 780 number). It's annoying you can't use css variables in media queries... Anyway, should I then replace all instances of 780px to 1136px?

zarSou9: Actually I lied, upon making that change and playing around with it, I realized that there actually is a reason for the two breakpoints. There is an in-between state - e.g. try moving the screen width on the [how-can-i-help page](https://aisafety.info/how-can-i-help).

LeMurphant (changes requested):
I'm seeing improvements on click vs hover around 1000px (half-ish screen on 1080p), but what is the expected behavior for the width of the popup? I'm still seeing a wide glossary popup at that resolution.

approx 1000px

zarSou9:
Oh whoops, tru dat, should be good now

LeMurphant:
The width is good, but I have this weird issue where, if an image from the glossary loads for the first time in mobile mode, it fails to load, and keeps failing even if I go to fullscreen. But I never get that issue for loads in fullscreen.

image_not_loading

Once the image has successfully loaded (in fullscreen) in this session, it shows up ok on mobile size.

To check if that was an issue previously, I did a checkout to main, and the issue did not show up then.

zarSou9:
Hmm, yeah that is weird. I noticed that as-well, but assumed it was a weird local issue, because I still saw it when checked out at master. But I removed the one line that might have been causing issues (which I realized wasn't needed). Feel free to try it now

LeMurphant (changes requested):
Unfortunately, the load issue is still present, does not seem to have changed since last commit.

zarSou9:
Alright well I wasn't able to find a working solution for the original issue. Which was that if the user starts out with a smaller screen, then widens it or vice versa, you get undesired behaviour. So this PR just refactors a bit now.

Issue #613, opened 2024-05-02
Remove spacing between bullet points and lists


LeMurphant:
![spacing](https://github.com/StampyAI/stampy-ui/assets/16197732/9890636b-07ce-41ae-b0a5-65423cdf285c)
https://docs.google.com/document/d/1YKA4i_3rfMMGrWLtvzNy9IplYgS4Ialdc6f-Rfu-CBg/edit
https://aisafety.info/questions/9B85/Isn't-the-real-concern-misuse

It seems to me (and Taylor and Steven) that the vertical spacing between bullet points (or numbered points) is excessive

bryceerobertson:
@LeMurphant is this looking better now?

LeMurphant:
Unfortunately we now have issues with insufficient spacing in some cases
![spacing](https://github.com/user-attachments/assets/e196c16b-2d4d-497f-bba7-17588c8ab89a)
![spacing_gdocs](https://github.com/user-attachments/assets/cac980f3-6485-4769-9fd4-12748dd104a9)

LeMurphant:
This might be related to https://github.com/StampyAI/stampy-ui/issues/352

Algon-33:
Claire, @LeMurphant's friend, wants there to be more space after embedded videos in our articles. This seems like a related problem. Please fix it.

LeMurphant:
There are also issues with increasing the space after sub-headings, e.g.
https://aisafety.info/questions/NM3E/13:-We-are-not-reliably-on-track-to-solving-these-problems

Algon-33:
This is relatively high priority because of how it makes the site feel.

LeMurphant:
@Algon-33 I agree. IIRC I diagnosed that the issue was in GDocsRelatedThings, I'll try to get to it before next Monday

Algon-33:
TY

LeMurphant:
The fix did not work, the rich text in Coda seems unchanged.

My guess is that the issue is that the parser proceeds differently than devtools.js, I will investigate

Algon-33:
Ah, that sucks.

PR #940, opened 2025-05-01
Add markdown debugging tool

LeMurphant:
While diagnosing https://github.com/StampyAI/stampy-ui/issues/613, I created a tool to view the markdown that is fetched from Coda to check for discrepancies with the output of GDocsRelatedThings (there were none).

It contains a static html page where you can insert your Coda token as well as a specific page ID. The Coda answer is then fetched and displayed, with a specific emphasis on the markdown.

This works well right now and I will use it, but I realize that as the repo changes, it might become out of date. For this reason, I'm not sure if it should be merged. Feel free to just reject this and I will keep a local copy.


debug-tools/README.md
Aprillion, line 1: hmm 🤔 I am not strongly against, but it would sound more useful to me if this was a route similar to `/cache`, re-using some of the existing code instead of duplicating a bunch of code that might get out of sync (whether or not it would be behind login reusing existing secretes or providing the key via input field doesn't matter as much)...

let's keep the PR open for now and if you find 1 more person who finds it immediatelly useful, I'd be fine with merging ... if it's just for you, I vote to keep it only on your localhost 😅

LeMurphant: Makes sense. I'll convert the PR as a draft (0 other people are actively investigating markdown issues) and eventually make something that uses a /route.

PR #942, opened 2025-05-09
Cache page improvements

LeMurphant:
Allows clearing of a single cache entry (e.g. banners).

Fixes the Cache page in development environments.

Pretty-prints values

Pre-fetches titles for article IDs (Algon's request)

### DIFF
--- a/app/routes/cache.tsx
+++ b/app/routes/cache.tsx
@@ -1,6 +1,6 @@
 import {ActionFunctionArgs, LoaderFunctionArgs, json} from '@remix-run/cloudflare'
 import {useLoaderData, useActionData, useNavigation, Form} from '@remix-run/react'
-import {useEffect, useState} from 'react'
+import React, {useEffect, useState} from 'react'
 import {isAuthorized} from '~/routesMapper'
 import {loadCacheKeys, loadCacheValue, cleanCache} from '~/server-utils/kv-cache'
 
@@ -20,7 +20,29 @@ export const loader = async ({request}: LoaderFunctionArgs) => {
     return json([] as string[], {status: 401})
   }
 
-  return await loadCacheKeys()
+  const keys = await loadCacheKeys()
+
+  // Pre-fetch titles for article-like keys (short keys with only alphanumeric chars)
+  const articleKeys = keys.filter((key) => /^[A-Z0-9]{4}$/.test(key))
+
+  // Load cached values and extract titles
+  const titles: Record<string, string> = {}
+
+  for (const key of articleKeys) {
+    try {
+      const value = await loadCacheValue(key)
+      if (value) {
+        const parsed = JSON.parse(value)
+        if (parsed && parsed.data && parsed.data.title) {
+          titles[key] = parsed.data.title
+        }
+      }
+    } catch (e) {
+      // Ignore errors for individual keys
+    }
+  }
+
+  return json({keys, titles})
 }
 
 export const action = async ({request}: ActionFunctionArgs) => {
@@ -64,7 +86,12 @@ export const action = async ({request}: ActionFunctionArgs) => {
 }
 
 export default function Cache() {
-  const keys = useLoaderData<typeof loader>()
+  const loaderData = useLoaderData<typeof loader>()
+  const keys = Array.isArray(loaderData) ? loaderData : loaderData.keys
+  const articleTitles: Record<string, string> = Array.isArray(loaderData)
+    ? {}
+    : loaderData.titles || {}
+
   const actionData = useActionData<typeof action>()
   const transition = useNavigation()
   // @ts-expect-error inferred type kinda looks OK, but TS is unhappy anyway
@@ -72,6 +99,7 @@ export default function Cache() {
   const [cacheValues, setCacheValues] = useState(() =>
     Object.fromEntries(keys.map((k) => [k, null]))
   )
+  // Article titles are pre-loaded by the server and never change during the session
   useEffect(() => {
     if (cacheKey) {
       setCacheValues((curr) => ({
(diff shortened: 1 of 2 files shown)


app/routes/cache.tsx
```diff
@@ -172,9 +172,27 @@ export default function Cache() {
                       backgroundColor: 'transparent',
                     }}
                   >
-                    {typeof cacheValues[key] === 'string'
-                      ? JSON.stringify(JSON.parse(cacheValues[key]), null, 2)
-                      : JSON.stringify(cacheValues[key], null, 2)}
+                    {(() => {
```
LeMurphant, line 175: Added this to address lint error
```
app/routes/cache.tsx:176:51 - error TS2345: Argument of type 'null' is not assignable to parameter of type 'string'.

176                       ? JSON.stringify(JSON.parse(cacheValues[key]), null, 2)
```
There might be a better way to do it

app/routes/cache.tsx
```diff
@@ -19,7 +20,29 @@ export const loader = async ({request}: LoaderFunctionArgs) => {
...
+    try {
+      const value = await loadCacheValue(key)
+      if (value) {
+        const parsed = JSON.parse(value)
+        if (parsed && parsed.data && parsed.data.title) {
+          titles[key] = parsed.data.title
+        }
+      }
+    } catch (e) {
+      // Ignore errors for individual keys
```
Aprillion, line 41: not sure if await inside for loop is a good idea, perhaps it's fine since this code runs on CF and waits from KV store, would need to give it a bit more time to think through..

but I don't agreee with using try/catch with empty catch 🤔

app/routes/cache.tsx
```diff
@@ -31,26 +54,52 @@ export const action = async ({request}: ActionFunctionArgs) => {
...
-  } else if (actionKey === Actions.loadCache) {
-    return json({cacheKey, cacheValue: await loadCacheValue(cacheKey)})
-  } else if (actionKey === Actions.reload) {
-    return null
-  } else {
-    return json({error: `Unknown action ${actionKey}`}, {status: 400})
+
+  try {
+    if (actionKey === Actions.delete) {
+      console.log('Cleaning all cache')
```
Aprillion, line 60: what's the deal with the console logs - do you know where in CF to find them? would you be checking them there?

app/routes/cache.tsx
```diff
@@ -31,26 +54,52 @@ export const action = async ({request}: ActionFunctionArgs) => {
...
+      console.log(`Clearing single cache key: ${cacheKey}`)
+      await STAMPY_KV.delete(cacheKey)
+      return json({message: `Successfully cleared cache for: ${cacheKey}`})
+    } else {
+      return json({error: `Unknown action ${actionKey}`}, {status: 400})
+    }
+  } catch (err) {
+    console.error(`Error processing ${actionKey} for ${cacheKey}:`, err)
+    const errorMessage = err instanceof Error ? err.message : 'Unknown error occurred'
+    return json({error: `Error processing cache action: ${errorMessage}`}, {status: 500})
```
Aprillion, line 84: what's the difference of this new code from normal HTTP 500 response if there was no try/catch?

app/routes/cache.tsx
```diff
@@ -67,17 +116,88 @@ export default function Cache() {
...
+                  {articleTitles[key] && (
+                    <span style={{fontWeight: 'normal', fontStyle: 'italic', marginLeft: '8px'}}>
+                      ({articleTitles[key]})
+                    </span>
+                  )}
+                </strong>
+              </div>
+            </li>
+            {cacheValues[key] && (
+              <li key={`${key}-value`} style={{marginTop: '-5px', marginBottom: '15px'}}>
```
Aprillion, line 155: (minor) we should keep only 1 `<li>` without fragment, these are not 2 separate list items.. the `<div>` should be enough inside conditional part

app/routes/cache.tsx
Aprillion: yeah, should be possible to write with simpler code .. but not a big deal, the meaning seems clear I guess

Aprillion (approved):
since it's dev-only page, feel free to merge now and fix whatever won't work as well as intended later

app/routes/cache.tsx
LeMurphant: The try/catch is because there could be false-positives in matching articleKeys, I wanted to ignore those. What would be a cleaner way to do this?

Since it's dev-only, I don't really care about speed, but if you think it's a problem I can just remove this title-fetching feature.

app/routes/cache.tsx
LeMurphant: I see them when I run `npm run dev` locally, but it makes sense that I should avoid clogging the logs just for that. I removed them.

app/routes/cache.tsx
LeMurphant: Does https://github.com/StampyAI/stampy-ui/pull/942/commits/607dd116f4335f4d5febd46743e625b81f7d75b7 address this properly?

app/routes/cache.tsx
LeMurphant: It was an AI suggestion that seemed to make sense, but the previous version worked fine without it, so I reverted it in https://github.com/StampyAI/stampy-ui/pull/942/commits/bd8b5844a7f67015149594ef8f7b3d8eb9df2949

app/routes/cache.tsx
Aprillion: yes, thank you

app/routes/cache.tsx
Aprillion: > there could be false-positives in matching articleKeys

if that is a bug, please report an issue (or what's the number of existing issue?) 🙏 
if not a bug and/or if it is about typos in GDocs and we already have some script to report the problems in Discord server or elsewhere, please add a note in code comments how the exceptions are handled elsewhere so we don't need to do anything about the false positives here

Aprillion (approved):

app/routes/cache.tsx
LeMurphant: I merged but we can still discuss this.

What I meant is that some keys (such as _allQuestions_ and _redirects_) do not correspond to article keys, and I am only interested in the latter. Filtering for `/^[A-Z0-9]{4}$/` mostly fixes that, but in theory I could imagine something like *FEAT* that is not an article, but that fits the regex, and the code is to exclude these. It might make more sense to check if these keys have a *data->title* attribute and ignore them if they don't instead of a try/catch though.

Since I merged, I notice that https://aisafety.info/cache takes 5+ seconds to load (it's instant on dev), and this is probably the cause. I think it's a good enough reason to remove this feature (titles with keys) entirely.

Aprillion: ah, I see, thanks for the explanation - code comments are ideal place to explain reasons *why* some code is there, e.g. `// Ignore errors for individual keys` is completely useless comment but `// Ignore errors for keys that don't match any articles, like allQuestions and redirects, but not filtered out by the above regex` would be perfect for this kind of documentation IMO

yeah, seems really slow in production and best to undo after all 😢

PR #943, opened 2025-05-10
Improve list paragraph spacing

LeMurphant:
Fix https://github.com/StampyAI/stampy-ui/issues/613

Addresses spacing of list items. I also shrank the spacing between lines inside list items, IMO it makes it more readable.

The best way to test this is to actually run it in the dev environment, but here are a few screenshots

https://aisafety.info/questions/8185/What-is-Goodhart's-law
goodhart

https://aisafety.info/questions/NM38/4:-The-road-from-human-level-to-superintelligent-AI-may-be-short
road_superintelligence

https://aisafety.info/questions/3485/What-are-accident-and-misuse-risks
misuse

### DIFF
--- a/app/components/Article/article.css
+++ b/app/components/Article/article.css
@@ -97,7 +97,6 @@ article .contents p + ol {
 /* Styling for list items */
 article .contents li {
   padding-bottom: var(--spacing-8);
-  line-height: 160%; /* Tighter line spacing within list items */
 }
 
 article .contents > p:has(> img) {
--- a/app/components/Article/article.css
+++ b/app/components/Article/article.css
@@ -77,19 +77,27 @@ article .contents a:visited {
 }
 
 article .contents p {
-  padding-bottom: var(--spacing-32);
+  padding-bottom: var(--spacing-24);
 }
 
 article .contents p:last-child {
   padding-bottom: 0;
 }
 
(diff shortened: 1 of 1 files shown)


bryceerobertson (approved):

app/components/Article/article.css
```diff
@@ -77,19 +77,27 @@ article .contents a:visited {
...
+/* If a list is preceded by a paragraph, reduce top margin */
+article .contents p + ul,
+article .contents p + ol {
+  margin-top: -16px;
+}
+
+/* Styling for list items */
+article .contents li {
+  padding-bottom: var(--spacing-8);
+  line-height: 160%; /* Tighter line spacing within list items */
```
LeMurphant, line 100: @bryceerobertson if I keep the original spacing between the lines in the paragraph, it looks like this

orig_spacing

My thought is that due to my reduced margins overall, it makes it hard to distinguish if we are still in a list or not. Feel free to run this in your dev environment and see if it looks better if you remove this line.

bryceerobertson: Yeah I personally prefer it with the line removed, but that's just my 2c

Aprillion (approved):
last commit looks OK to me if blocked by waiting for re-review

Issue #946, opened 2025-05-14
Increase spacing between headers and paragraphs


Algon-33:
The spacing between headers and text looks ugly. Add some spacing to the headers. Perhaps add to the CSS style "  margin-bottom: 30px;" for all headers. Or using padding-bottom. 

![Image](https://github.com/user-attachments/assets/92cfdabe-5085-4c02-bed5-cf94c202da23)

PR #947, opened 2025-05-14
Add spacing under headers

LeMurphant:
Fix https://github.com/StampyAI/stampy-ui/issues/946


Algon-33 (approved):
Visually, this is a clear improvement over what we had before.

Issue #931, opened 2025-04-10
Carousel load times are long


Algon-33:
When switching between videos in the carousel, it takes about 1-2 seconds for the new video to load. In the meanwhile, there's just a black page. This looks bad and feels jarring. It would be good if we could either reduce the load times or do something to fix the jarring change. E.g. cache the image that is shown when the video is loaded (or grab it from the video source somehow? Maybe it's the thumbnail?) and display it while the video is loading.

PR #954, opened 2025-05-24
Preload adjacent MediaCarousel images

LeMurphant:
Fix https://github.com/StampyAI/stampy-ui/issues/931

Solves slow loading of images, including static images for video media.

I could not get the actual video to load faster, although IMO it's fine.

Roping in more experienced front-end developers to check if my LLM-fueled code contains aberrations.


app/components/MediaCarousel/index.tsx
```diff
@@ -37,10 +37,43 @@ type MediaCarouselProps = {
...
+        img.src = item.url
+      }
+    })
+  }, [adjacentIndices, items])
+
   return (
     <div className="media-carousel-container padding-bottom-32">
       <div className="media-carousel-track">
-        <Media item={items[currentIndex]} />
+        {/* Render current and adjacent items but only show current */}
```
Aprillion, line 61: why? this code should work, but AFAICT rendering empty hidden divs won't help in any way, the above preloading should work even if you keep `<Media item={items[currentIndex]} />` like before

Aprillion (dismissed):

app/components/MediaCarousel/index.tsx
```diff
@@ -37,10 +37,43 @@ type MediaCarouselProps = {
 const MediaCarousel = ({items}: MediaCarouselProps) => {
   const [currentIndex, setCurrentIndex] = useState(0)
 
+  // Calculate adjacent indices
+  const adjacentIndices = useMemo(() => {
+    const prevIndex = currentIndex - 1 < 0 ? items.length - 1 : currentIndex - 1
+    const nextIndex = currentIndex + 1 >= items.length ? 0 : currentIndex + 1
+    return new Set([prevIndex, currentIndex, nextIndex])
```
zarSou9, line 44: If you get rid of the empty div preloading, currentIndex is redundant. Also it might be simpler for this to just return nextIndex.

app/components/MediaCarousel/index.tsx
LeMurphant: I removed it, and performance (in Chrome) went back to baseline. Since your comment made sense, I asked the LLM why that was, and it answered this:
```
While Aprillion's comment was technically correct that new Image() preloading should work, in practice the DOM-based approach provides better performance because:

  1. Priority: DOM elements get higher loading priority than JavaScript-created images
  2. Browser optimization: Modern browsers are optimized to handle hidden DOM elements efficiently
  3. Immediate loading: The browser starts loading resources as soon as they're in the DOM, even if hidden
  ```

File: app/components/MediaCarousel/index.tsx (hunk 2 above)
Comment (LeMurphant) [reply app/components/MediaCarousel/index.tsx:44]: Just having the next index would indeed give us 80% of the benefits, but is the price of loading so high that we want to avoid loading the previous index? I'm thinking of when users click on "back" instead of forward.

File: app/components/MediaCarousel/index.tsx [hunk 3]
```diff
@@ -37,10 +37,43 @@ type MediaCarouselProps = {
...
+
+          return (
+            <div
+              key={index}
+              style={{
+                display: index === currentIndex ? 'block' : 'none',
+                position: index === currentIndex ? 'relative' : 'absolute',
+              }}
+            >
+              {isAdjacent && <Media item={item} />}
```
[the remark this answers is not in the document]
LeMurphant: Since I don't know about rendering, this is basically the LLM talking through me:
```
Here's the flow that updates adjacentIndices:

  1. User clicks navigation arrows (lines 84-85):
    - Left arrow: setCurrentIndex((i) => (i <= 0 ? items.length - 1 : i - 1))
    - Right arrow: setCurrentIndex((i) => (i >= items.length - 1 ? 0 : i + 1))
  2. State update triggers re-render:
    - When setCurrentIndex is called, currentIndex state changes
    - This causes the component to re-render
  3. useMemo recalculates (lines 41-45):
    - Since currentIndex is in the dependency array [currentIndex, items.length]
    - React detects that currentIndex changed
    - The function inside useMemo runs again, calculating new adjacent indices
   ```
   
 I added some debug code and tested on the 6-carousel at [this page](https://aisafety.info/questions/0), and my browser confirms that when I click on the arrows, the indexes get updated
  ```
  ➡️ Right navigation clicked
index.tsx:47 🔄 useMemo recalculating adjacentIndices
index.tsx:48   currentIndex: 5
index.tsx:52   new adjacentIndices: (3) [4, 5, 0]
````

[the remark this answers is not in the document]
zarSou9: Gah, yeah don't listen to me, I got confused by how useMemo works. It still updates the value, just doesn't recalculate it, i.e. it memoizes values 🤦‍♂️. I got confused, thinking that the motivation of useMemo was to only update the adjacent items once. Yeah this works 👍

PR #963, opened 2025-07-13
Stampy chat 47 filter non published related questions

jrhender:
This PR filters the `relatedQuestions` used to show suggestions after a human written article is used as a message in the stampy chat history. Fixes: https://github.com/StampyAI/stampy-chat/issues/147

The filtering is applied to generally to the `loadQuestionDetail` function in `stampy.ts`, rather than using client side filtering. The rationale is that this filtering should probably always be applied and it would be better to avoid a similar future bug where relatedQuestion filtering is not avoid to a use case in the client code.

#### "Warning shot" test case
When entering "warning shot" as the first message in the chat, the unpublished articles `Should we expect "warning shots" before an unrecoverable catastrophe?` and `Would "warning shots" make a difference and, if so, would they be helpful or harmful?` are not displayed as suggestions

Status on coda
image

Currently on live site:
image

Local test with PR change:
image

### DIFF
--- a/app/server-utils/stampy.ts
+++ b/app/server-utils/stampy.ts
@@ -355,12 +355,14 @@ export const loadQuestionDetail = withCache('questionDetail', async (question: s
   )) as AnswersRow[]
   const questionData = convertToQuestion(rows[0])
 
-  const {data: onSiteAnswers} = await loadOnSiteAnswers('NEVER_RELOAD')
-  const onSitePageIds = new Set(onSiteAnswers.map((q) => q.pageid))
+  const {data: allQuestions} = await loadAllQuestions('NEVER_RELOAD')
+  const viewablePageIds = new Set(allQuestions.filter(isQuestionViewable).map((q) => q.pageid))
 
   return {
     ...questionData,
-    relatedQuestions: questionData.relatedQuestions.filter(({pageid}) => onSitePageIds.has(pageid)),
+    relatedQuestions: questionData.relatedQuestions.filter(({pageid}) =>
+      viewablePageIds.has(pageid)
+    ),
   }
 })
 


app/components/Chatbot/index.tsx
```diff
@@ -237,11 +237,34 @@ export const Chatbot = ({question, questions, settings}: ChatbotProps) => {
...
-            (question.relatedQuestions || [])
-              .slice(0, 3)
-              .map(({title, pageid}) => ({text: title, pageid}))
-          )
+          const filteredFollowups = (question.relatedQuestions || [])
+            .filter(({pageid}) => {
+              const q = questions?.find((q) => q.pageid === pageid)
+              if (q?.status === undefined) return false
+              switch (q.status) {
+                case QuestionStatus.LIVE_ON_SITE:
```
mruwnik, line 245: why have a full blown switch for the equivalent of a `q?.status === QuestionStatus.LIVE_ON_SITE`? I suppose a move generic question would be why is the articles endpoint not doing this filtering for you? Is there any place where you'd want non live on site related questions to be shown? If not, then I'd do this when transforming the coda responses. Though that requires all questions to be loaded by then, which is a sort of chicken and egg situation :/

LeMurphant:
The chatbot currently crashed for all non-human written queries, both live and on this branch.
gone_wrong
This makes it hard to investigate this issue
@mruwnik could this be related to some of the changes you did recently?

mruwnik:
should be back up now

app/components/Chatbot/index.tsx
```diff
@@ -237,11 +237,34 @@ export const Chatbot = ({question, questions, settings}: ChatbotProps) => {
...
           // question.relatedQuestions = question.relatedQuestions.slice(0,2);
-          setFollowups(
-            (question.relatedQuestions || [])
-              .slice(0, 3)
-              .map(({title, pageid}) => ({text: title, pageid}))
-          )
+          const filteredFollowups = (question.relatedQuestions || [])
+            .filter(({pageid}) => {
+              const q = questions?.find((q) => q.pageid === pageid)
+              if (q?.status === undefined) return false
```
LeMurphant, line 243: I got "What is a "warning shot"?" to recommend itself, can we add a check about the ID not being itself? It's fine if it loops, e.g. A->B->A

jrhender:
Changed to draft because I think the filtering in the current code is being done [on a subset of random questions](https://github.com/StampyAI/stampy-ui/blob/e6885d5c1321b604c58503efbf8763ff693e19cd/app/hooks/useOnSiteQuestions.ts#L37), not the full list of questions on the site, so it would exclude many follow up questions unnecessarily.

app/components/Chatbot/index.tsx
mruwnik: yes, that would be better if it's a simple change

jrhender: @mruwnik unfortunately, I don't think the "additional property in relations objects" is possible from the Coda API.
- `Related Answers DO NOT EDIT` seems to be a [relation column](https://help.coda.io/en/articles/1385997-connect-tables-with-relation-columns)
- I think that the API is returning a generic set of data as all of the current properties (`@context`, `@type`, `additionalType`, etc) seem like they are generic to any coda relation. 
- Coda [doesn't support](https://community.coda.io/t/api-how-to-query-for-a-relation/52025) querying using relation column properties, which IMO suggests that including additional properties in relational object results is probably not supported as they are related functionality

I'll go back to this approach:
> adding a 2nd network request as a query for the related questions' status, after the 1st network request for the main question's details.

With caching, I think this approach wouldn't incur too great a latency cost.

jrhender: @mruwnik I've implemented the approach using a 2nd network request for the related questions' status. https://github.com/StampyAI/stampy-ui/pull/995 Do you think it makes sense to continue with the approach?

app/routes/questions.$questionId.$.tsx
jrhender: Code removed from this page as the `relatedQuestions` are now filtered in the server-utils code

PR #969, opened 2025-08-06
Update dependancies

LeMurphant:
Written by Aprillion


LeMurphant:
@Aprillion In the process of updating, I ran into issues. Claude says:

`The main issue is actually just React 19 vs React 18. Since your package.json has React 19 but Remix v2.17 expects React 18, that's why we'd need --legacy-peer-deps.`

I used that flag, and it seems to work. I have a warning that
`The CJS build of Vite's Node API is deprecated. See https://vite.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details`
I don't know if this is important.

I did a quick scan of the website, and found that the Caroussel under the word "months" does not appear in
http://localhost:8787/questions/NM3T/
Surprisingly, it does appear under
http://localhost:8787/questions/NM37/1:-AI-is-advancing-fast
Maybe it's the video breaking it?

Aprillion:
@LeMurphant warning about CJS doesn't sound important and no idea about why Caroussel might only work "sometimes"

however `npm ci` doesn't seem to work without the `--legacy-peer-deps` in https://github.com/StampyAI/stampy-ui/actions/runs/16786489709/job/47538173650?pr=969 .. not sure we should add it to the build step, it doesn't feel right even though I can't think of any concrete reason why not .. perhaps sticking to React 18 might be better if Remix doesn't support 19 🤔

LeMurphant:
> no, I did it in my branch while investigating the caching issue (to make sure it's not a React bug .. it was not), so feel free to drop major updates

I reverted to 18.3.1 in https://github.com/StampyAI/stampy-ui/pull/969/commits/40fc6036cede3cd219c443e4a6576478bfc0354a

LeMurphant:
I tested it for ~30m, can't find anything broken.

Aprillion (approved):
LGTM 🤞

Issue #841, opened 2024-11-07
Advanced section: change AI images to icons


bryceerobertson:
From Murphant:
To change it, just update the "Icon" column in this table. It might take 30-minutes-ish to update
https://coda.io/d/AI-Safety-Info_dfau7sl2hmG/All-Answers_suL3ndPS#_luXkqiO0

bryceerobertson:
@melissasamworth let me know when the icons are done and I'll throw them in

PR #970, opened 2025-08-07
Remove images for Advanced sections on Frontpage

LeMurphant:
fix https://github.com/StampyAI/stampy-ui/issues/841

I just removed the images


Algon-33 (approved):
Looks good to me.

PR #971, opened 2025-08-07
Fix chat bot

mruwnik:
* Fixes bugs causing chat to break
* Potentially better handling of duplicate references
image

* Adds block for thinking if available
image


app/components/Chatbot/ChatEntry.tsx
```diff
@@ -263,7 +308,11 @@ const ChatbotReply = ({
           ]}
         />
       ) : undefined}
-      {phase === 'followups' ? <p className="followups">Checking for followups...</p> : undefined}
+      {phase === 'followups' ? (
+        <p className="followups">
+          Checking for followups<span className="animated-ellipsis"></span>
```
nerualauren, line 313: I don't like using a fake animation here. in the playground I used an animation which is progressed by receiving more thinking tokens, I'm planning to port that over.

nerualauren (approved):
looks reasonable to me, I didn't fully step through all of the code paths in my head.

in an upcoming pr, I'll be adding text fragments to the citation urls, we'll need to decide if it makes sense to dedup links with different fragments.

app/components/Chatbot/ChatEntry.tsx
mruwnik: dunno about that... It depends what exactly you want to communicate. Having it animated per chunk makes sense if you know what it's doing, but if you don't, then it just looks janky? Might be worth asking @melissasamworth to chime in

Issue #229, opened 2023-05-07
Feature Request : Light/Dark Mode toggle button


markovial:
A button that switches between light and dark mode on the website. [Here is a link](https://wordpress-theme.spider-themes.net/ama/home-classic/) preview to a sample website. Below are some screenshots of something like what I had in mind, the toggle button is on the top right.

![image](https://user-images.githubusercontent.com/52564817/236705002-6ea07f9a-4ad2-4f6e-9ee3-42e3c89b6f8c.png)
![image](https://user-images.githubusercontent.com/52564817/236705009-e08879ad-aefd-40f1-ac97-f3814a5734d7.png)

Aprillion:
Will be possible to implement using `const {savedTheme, setStorageTheme} useTheme()`hook from #307 and see whatever new design will be decided in #285.

bryceerobertson:
An MVP could be:
Add a button that links people to download a Chrome extension like [this one](https://chromewebstore.google.com/detail/dark-mode/declgfomkjdohhjbcfemjklfebflhefl?hl=en).

PR #972, opened 2025-08-08
Dark mode

LeMurphant:
Fix https://github.com/StampyAI/stampy-ui/issues/229

image

This is 90% vibe coded. Written with Claude4.1, critiqued with GPT-5 (Cursor). I did a single pass to remove what I thought was superfluous, but I expect to have missed some.

This is a MVP, I expect some small things to break, but this seems better than the status quo.

I initially wanted to remove the white bleeding, but Algon said he likes it, open to other opinions.


LeMurphant:
Lauren and Steven prefer the darker color found in commit 395e78ef2f3a5b4762b684d19385c938baa24071

mruwnik (commented):
I'm sure @melissasamworth will have Thoughts, but I'm all for a dark mode

app/root.css
```diff
@@ -28,6 +28,15 @@
 
   --colors-green-500: #40c075;
 
+  /* Dark mode semantic colors */
+  --background: #ffffff;
```
Aprillion, line 32: wut? I guess comment can be deleted, it probably tried to explain something but it ended up confusing/useless

app/root.css
```diff
@@ -52,6 +61,20 @@
...
+  --background: #2d2d2d;
+  --text: #e8e8e8;
+  --text-secondary: #a8a8a8;
+  --border: #404040;
+  --card-background: #3a3a3a;
+  --hover-background: #1a3330;
+  --selected-background: #4a4a4a;
+
+  /* Keep utility colors consistent */
+  --colors-white: #ffffff;
```
Aprillion, line 75: this line is not doing anything...

app/root.css
```diff
@@ -917,15 +945,15 @@ button {
     position: fixed;
     top: 0;
     left: 0;
-    background-color: #1b2b3e99;
+    background-color: rgba(27, 43, 62, 0.6);
```
Aprillion, line 948: no `var` here? if on purpose, please add comment why not

app/components/Article/article.css
```diff
@@ -138,7 +138,7 @@ article .contents .link-popup img {
 }
 
 article .glossary-entry {
-  border-bottom: 3px dotted var(--colors-cool-grey-500);
+  border-bottom: 3px dotted var(--text-secondary);
```
Aprillion, line 141: the color was different previously, is the change on purpose?

app/components/Article/article.css
```diff
@@ -177,6 +177,7 @@ article .contents a.button {
 }
 article .link-popup .glossary-popup > .contents {
   padding: var(--spacing-24) var(--spacing-40) var(--spacing-24);
+  color: var(--text);
```
Aprillion, line 180: why is this needed?

app/components/Article/article.css
```diff
@@ -203,6 +204,7 @@ article .link-popup.shown {
 }
 article .link-popup p {
   margin: 0;
+  color: var(--text);
```
Aprillion, line 207: why is this needed?

app/components/Chatbot/widgit.css
```diff
@@ -29,7 +29,7 @@
 }
 
 .red {
-  color: #d40000;
+  color: #ff4444;
```
Aprillion, line 32: ?

app/components/Feedback/feedback.css
```diff
@@ -27,6 +27,8 @@
   width: 384px;
   padding: var(--spacing-32);
   border-radius: var(--border-radius);
+  background: var(--card-background);
+  border: 1px solid var(--border);
```
Aprillion, line 31: how is a new border related to dark mode?

Aprillion (dismissed):

app/components/Chatbot/widgit.css
LeMurphant: It's a change in the color which GPT-5 argued is more appropriate for light and dark backgrounds simultaneously. 
Original:
original_red
New:
new_red

I slightly agree new is better, but ok to drop it if this change is not appropriate.

LeMurphant:
> I feel like the buttons (e.g. Click here) should be a different colour from the background

I'd rather somebody with some experience take all comments on colors (and layout) and update the look while following a plan, instead of me making piecemeal changes. Another option is to have a meeting where interested people prototype something and we agree on a layout.

For now I want to emulate as closely as possible the light mode and only address layout issues that give a _broken_ feel.

app/root.css
LeMurphant: The change was just for readabilty (GPT-5 suggested rgba > hex), but I ended up using a var instead.

app/components/ThemeToggle/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+    (mounted && window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light')
+
+  const isDark = currentTheme === 'dark'
+
+  const toggleTheme = () => {
+    const newTheme = isDark ? 'light' : 'dark'
+    setStorageTheme(newTheme)
+  }
+
+  // Don't render until mounted to avoid SSR mismatch
```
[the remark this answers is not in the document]
LeMurphant: I could not generate an SSR mismatch even without that code, so I removed it.

app/components/Chatbot/widgit.css
Aprillion: no opinion on color as long as the change is on purpose and not accidental..

Issue #911, opened 2025-03-14
Browser hides text while loading fonts.


Algon-33:
Our text font sometimes takes a while to load in, causing the browser to hide the text until the font loads. Which causes a flash of invisible text (FOIT). This isn’t great for users. The easiest fix is to use font-display: swap in your @font-face style. This will solve the  FOIT issue for most modern browsers. 

The page which raised this issue to awareness was https://aisafety.info/questions/5611/Why-can%E2%80%99t-we-just-turn-the-AI-off-if-it-starts-to-misbehave.

Issue #900, opened 2025-03-14
AIsafety.info/chat doesn't have an h1 heading.


Algon-33:
Add a concise, relevant h1 heading to improve SEO for this page. Maybe "Stampy "or "Chat" or so on.

Issue #906, opened 2025-03-14
Unused JS slowing down some pages.


Algon-33:
Unused JavaScript is present on the site. This slows down pages. It would be good if we could remove such unnecessary code from pages. As some potential examples of pages which have unnecessary JS, SEMRush’s crawler flagged the following pages.

https://aisafety.info/questions/6412/What-about-technological-unemployment-from-AI 
https://aisafety.info/
https://aisafety.info/questions/9OGZ/
https://aisafety.info/chat/
https://aisafety.info/how-can-i-help
https://aisafety.info/questions/8E41/Will-AI-be-able-to-think-faster-than-humans
https://aisafety.info/questions/5611/Why-can%E2%80%99t-we-just-turn-the-AI-off-if-it-starts-to-misbehave
https://aisafety.info/questions/9TDI/Objections-and-responses
https://aisafety.info/categories/

Please investigate these pages to see if there’s any common thread in the JS that’s unnecessary, and if so, remove it.

LeMurphant:
Claude:
`  I've successfully optimized the JavaScript bundle size by removing unused code and dependencies. Here are the changes made to address issue #906:

  Removed Unused Dependencies:

  1. copy-to-clipboard - Not used anywhere (the Article component uses native clipboard API)
  2. isbot - Not imported or used anywhere in the codebase

  Removed Unused Code:

  1. Deleted copyLink.tsx component - Not imported anywhere
  2. Removed unnecessary CSS imports - Removed widgit.css imports from questions.allQuestions.ts and editors.tsx where they weren't needed

  Performance Optimizations:

  1. Lazy-loaded client-zip - Changed to dynamic import so it's only loaded when actually downloading files, not on every page load

  These optimizations will reduce the JavaScript bundle size across all pages, improving initial page load times and reducing unnecessary network transfers. The changes are particularly beneficial for:
  - Homepage (doesn't need zip functionality)
  - Question pages (don't need chatbot CSS or zip functionality)
  - Categories page (doesn't need zip functionality)

  The removed dependencies and lazy-loading will result in smaller bundle sizes and faster page loads across the entire site.
`

PR #973, opened 2025-08-09
SEO Improvements

LeMurphant:
Fix https://github.com/StampyAI/stampy-ui/issues/900
Fix https://github.com/StampyAI/stampy-ui/issues/911
Hopefully improve https://github.com/StampyAI/stampy-ui/issues/906


app/components/copyLink.tsx
```diff
@@ -1,34 +1,24 @@
...
-    <a
-      href={to}
-      className={`icon-link share ${copied ? 'copied' : ''}`}
-      onClick={shareLink}
-      {...props}
-    >
-      {children}
-    </a>
+    <button onClick={handleClick} {...props}>
+      {copied ? 'Copied!' : children}
```
Aprillion, line 21: the change from using copied class to explicit text is part of SEO improvements?

app/root.tsx
```diff
@@ -83,10 +83,23 @@ export const meta: MetaFunction<typeof loader> = ({data = {} as any}) => {
   ]
 }
 
-export const links: LinksFunction = () =>
-  [newStyles, cssBundleHref]
-    .filter((i) => i)
```
Aprillion, line 88: what the hell is happening here? is this old `filter` code useful for something, do we need to keep it? (could you dig in git history if you will have time, please? 🙏 ) ... if not sure and don't have time to investigate, feel free to keep it for now

Aprillion (approved):

app/components/Chatbot/index.tsx
```diff
@@ -195,8 +195,9 @@ const SplashScreen = ({
   <div className="stampy-widget padding-top-40">
     <IconStampyLarge />
     <div className="col-6 padding-bottom-40 padding-top-40">
-      <h2 className="teal-500">Hi there, I’m Stampy.</h2>
-      <h2>I can answer your questions about AI safety.</h2>
+      <h1 className="teal-500">Stampy Chatbot</h1>
```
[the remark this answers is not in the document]
LeMurphant: I didn't think of the possibility of creating the H1 for bots only, is that preferable? I have no opinion on whether the human-readable H1 is an improvement.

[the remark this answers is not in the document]
Aprillion: I meant more like `<h1>I can answer your question about AI safety</h1>` is possible and set CSS so nothing changes visually 🤔 but yeah, if we want text "Stampy Chatbot" for bots only, we could hide it in CSS, though search engines in particular would punish the site for using hidden h1, so that would beat the point of this SEO PR I guess 🙈 if no one else complains about too many headings, it's fine to keep it visible

app/root.tsx
Aprillion: yeah, no idea what's the `?url` import in remix and/or why gpt 5 think any other changed besides removing that one line with `.filter((i) => i)` are needed, without laptop right now and I don't think I want to understand it so much to search old docs on a phone 🙈 not important for this PR in any case

Aprillion (approved):

Issue #768, opened 2024-09-25
Glossary image takes up too much space


LeMurphant:
When an image is present, it often crops the text too heavily
![with_img](https://github.com/user-attachments/assets/d0d1163f-adff-4dce-849d-5e845994dbfb)
![no_img](https://github.com/user-attachments/assets/d835068d-2251-45db-a72c-eb20842dc9dc)
Making the popups deeper would not work since there's just not enough horizontal space to show a reasonable amount of text.

I see a few options:
1. Make all the glossary entries wider
2. Detect if entries have images, and if so make them wider
3. Automatically adjust the width so that the text fits.

LeMurphant:
@melissasamworth what do you suggest as a solution?

LeMurphant:
Fixed by https://github.com/StampyAI/stampy-ui/pull/986

PR #976, opened 2025-08-12
Move glossary popup to the right

LeMurphant:
Fix https://github.com/StampyAI/stampy-ui/issues/768, but also adds different features.

For large screens (~1500px wide), move glossary/footnotes popup to the right. This allows a bunch of nice options, like uncropped images.

Also changed logic for showing/hiding popups to reduce popup lag and shrank popup padding. Images are not a good indicator, I suggest trying it out.
image
image
image
Reducing width of window
image
image

90% vibe coded to get a feel for it. If we are happy with the result, I will go over the code more closely.


bryceerobertson:
I don't like the image being outside the box. Since Melissa's design had them in the box I think it's best to keep that

Issue #977, opened 2025-08-13
Could not resolve "react-router-dom"


Georann:
When trying to run the local dev environment, I got errors about the package `react-router-dom` missing. The package is imported from `app/components/Article/KeepGoing/index.tsx:21:28` and `app/routes/questions.$questionId.$.tsx:23:45`. And it does appear like it isn't included in the `package.json`. Am I missing something?

Georann:
@LeMurphant I'm getting the same error for `@cloudflare/kv-asset-handler` now

Georann:
Okay, it was my mistake that I tried to use pnpm

PR #978, opened 2025-08-13
Update dependancies

LeMurphant:
fix https://github.com/StampyAI/stampy-ui/issues/977

It looks like react-router-dom was imported by react by default, but was not under certain settings, so including it explicitly is better. I took this opportunity to check for other packages in a similar situation.

Not sure if this was the best way to do so, feel free to correct me.


Aprillion (approved):
not sure what are the repro steps for "certain settings" but the change shouldn't trigger other bugs, so if it helps, seems ok

Issue #980, opened 2025-08-18
Donation button up top


Algon-33:
Could we make a new button next to articles, how can I help and AI safety chatbot for donations? Rob's making a video mentioning us needing donations and will point people to our page. Here's the link to our fundraiser. https://www.every.org/aisafetyinfo/f/keep-stampy-alive

PR #981, opened 2025-08-18
Algon menu donate button 980

Algon-33:
Add a donate button on the menu at the top of the page, as per issue 980. https://github.com/StampyAI/stampy-ui/issues/980


app/components/icons-generated/Dollar.tsx
LeMurphant, line 1: Where is this used?

LeMurphant (changes requested):
DonateHeart seems to be the only icon that is used, why are the others present?

Functionally, it looks good. I thought maybe opening in a new tab would be preferable, but I'm not sure.

app/components/icons-generated/index.ts
```diff
@@ -63,6 +64,8 @@ export {default as Share} from './Share'
 export {default as Speaker} from './Speaker'
 export {default as Stamp} from './Stamp'
 export {default as Stampy} from './Stampy'
+export {default as StampyLarge} from './StampyLarge'
+export {default as StampySmall} from './StampySmall'
```
LeMurphant, line 68: The positions of these lines flip-flopping is annoying, can you follow these instructions and see if it solves it?
```
  To fix this, all contributors should:
  1. Run npm ci (instead of npm install) to install exact versions from package-lock.json
  2. Ensure they're using the Node version specified in engines (>=18)
  3. Regenerate icons after cleaning: npm run generate-icons
```

Aprillion (dismissed):

app/components/Nav/index.tsx
```diff
@@ -44,6 +45,12 @@ export const Nav = ({toc, categories}: NavProps) => {
             text="AI Safety Chatbot"
             reload
           />
+          <MenuItem
+            primary={true}
+            link="https://www.every.org/aisafetyinfo/f/keep-stampy-alive#/donate/card"
+            icon={<DonateHeart />} // Temporary - replace with donate icon later
```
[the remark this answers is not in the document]
Algon-33: Ah, my bad, I forgot to delete that comment.

app/components/icons-generated/index.ts
Aprillion: ah, the source of the problem seems to be that these 2 files are CamelCase instead of the correct dash-case and some operating systems don't distinguish case in file names while others sort them ASCII-bethically (uppercase first) => renaming the 2 offending files to lower-case would solve the problem

Algon-33: Changed StampyLarge and StampySmall in `app/assets/icons/` to dash-case.

Aprillion (approved):

app/components/icons-generated/index.ts
LeMurphant: Oh, I thought it was something about node versions since I first saw it appear in a PR @Aprillion made in node changes. This explanation makes sense, and I expect the change to dash-case to work.

LeMurphant (approved):
Was not able to test most recent changes, but based on the diff it looks good.

PR #982, opened 2025-08-21
Added donation button to menu mobile version

Algon-33:
Realized there wasn't a donation button in the menu of the mobile version of the site. It isn't ideal, because it's hidden in the menu, so many users will miss it. Perhaps we want a banner instead?


Algon-33:
Also, I just realized that on Chrome, the donate button looks pretty ugly: image


This is true in both the mobile and desktop version. I'll check what's gone wrong there.

PR #985, opened 2025-08-25
Bryce menu donate button 980

Algon-33:
Bryce asked us to change the icon for the donation button from a hand+heart to a heart. It looks good in my view. 
image
I recommend we go ahead with it.


LeMurphant (approved):
Code wise, looks good. I would prefer if @bryceerobertson can give his thumbs up before we merge.

PR #986, opened 2025-08-25
Popup images wikipedia style

LeMurphant:
Alternate to https://github.com/StampyAI/stampy-ui/pull/976

Follow guidelines in https://github.com/StampyAI/stampy-ui/pull/976#issuecomment-3206743392

Closer to original style

It would be cleaner to have the layout calculated with columns/rows, but I never got it to work properly.

### DIFF
--- a/app/components/Article/Contents.tsx
+++ b/app/components/Article/Contents.tsx
@@ -25,15 +25,17 @@ const footnoteHTML = (el: HTMLDivElement, e: HTMLAnchorElement): string | null =
   return elem.firstElementChild?.innerHTML || null
 }
 
+// TODO: Potential memory leak - event listeners are not cleaned up.
+// Consider refactoring to React components with proper cleanup in useEffect.
 const addPopup = (
   e: HTMLElement,
   id: string,
   contents: string,
   mobile: boolean,
   layout?: string
-): HTMLElement => {
+): void => {
   const preexisting = document.getElementById(id)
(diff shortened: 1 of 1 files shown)


LeMurphant:
> Images are still broken for me so can't evaluate properly

My bad @bryceerobertson , should be fixed now, at least in production. If you are having issues in dev, you can clear cache in http://localhost:8787/cache

bryceerobertson (approved):

LeMurphant:
I investigated Wikipedia's code, and they feed the dimensions of the image in the metadata at the same time the glossary info is fed. Updated this in https://github.com/StampyAI/GDocsRelatedThings/pull/101, we should merge both together.

bryceerobertson (dismissed):

app/components/Article/article.css
```diff
@@ -155,7 +155,90 @@ article .glossary-entry {
...
+/* Top image layout - image on top */
+article .glossary-popup.top-image-layout {
+  display: flex;
+  flex-direction: column;
+}
+
+/* Right image layout - image container */
+article .glossary-popup.right-image-layout .image-container {
+  flex: 0 0 180px;
+  order: 2;
```
Aprillion, line 179: huh?

LeMurphant:
I had to rewrite git history to sidestep merge issues. Now working on fixing issues that were raised.

Issue #993, opened 2025-08-27
Don't italicize glossary entries on italicized words


srkaas:
https://aisafety.info/questions/NM3T/A-case-for-AI-safety

"Superintelligent" is in italics and that's making the glossary entry on it display in italics, but it should be in normal text.

Issue #992, opened 2025-08-27
Show only one glossary entry


srkaas:
https://aisafety.info/questions/9J1L/Alignment-research

When I hover on "agent", it shows simultaneous popups for both "agent" and "agent foundations". It should show only one popup, probably the one for "agent foundations".

PR #994, opened 2025-08-28
Glossary detection and formatting

LeMurphant:
Fix https://github.com/StampyAI/stampy-ui/issues/993

Fix https://github.com/StampyAI/stampy-ui/issues/992

Also fixes some issues with terms being glossaried twice in the same page


app/components/Article/Contents.tsx
```diff
@@ -9,6 +9,21 @@ import {createRoot} from 'react-dom/client'
...
+  currentPageId: null as string | null,
+}
+
+const getOrResetSeenSet = (pageid: string): Set<string> => {
+  if (pageGlossaryState.currentPageId !== pageid) {
+    pageGlossaryState.seen = new Set<string>()
+    pageGlossaryState.currentPageId = pageid
+  }
+  return pageGlossaryState.seen
+}
```
jrhender, line 25: I'm a bit skeptical this global pageGlossaryState approach though it might be okay in practice. A problem that I see is that there is an assumption that all `Contents.tsx` rendered will be for the same page at any given time.

One alternative I think would make sense is to lift the `SeenState` to a parent component so that any child Contents components can share a `SeenState`. These are the React docs for this technique (just so that we're on the same page of which technique I have in mind): https://react.dev/learn/sharing-state-between-components. Not 100% this is the best idea, but it would allow for grouping Contents components in the parent component.

app/components/Article/Contents.tsx
```diff
@@ -96,28 +111,45 @@ const normalizeForComparison = (text: string): string => {
     .replace(/[‒–—]/g, '-') // Replace figure-dash, en-dash and em-dash with hyphen
 }
 
-const glossaryInjecter = (pageid: string, glossary: Glossary) => {
-  const seen = new Set()
+const glossaryInjecter = (pageid: string, glossary: Glossary, seen: Set<string>) => {
   return (html: string) => {
-    // Normalize for comparison to find terms regardless of formatting
-    const normalizedHtml = normalizeForComparison(html)
```
jrhender, line 103: Was it intentional to remove the normalization of the `html`?

jrhender (commented):
A couple thoughts from my end

app/components/Article/Contents.tsx
jrhender: Nice, yea, to me this new approach is better because it's more clearly and robustly ensuring that glossary entries will be unique within an `Article`. That said, I am confused about how terms were being glossaried twice prior to this fix because it seems each Article already only had [a single Contents](https://github.com/StampyAI/stampy-ui/blob/9433f390c208c92cf71ee6cbbd84b936294025cc/app/components/Article/index.tsx#L167-L174). So the local `seen` tracking in `Contents` that was there before should have been equivalent.

What were the reproduction steps for this "double glossaried" issue prior to this fix?

app/components/Article/index.tsx
```diff
@@ -154,6 +154,8 @@ type ArticleProps = {
 }
 export const Article = ({question, glossary, className, showNext}: ArticleProps) => {
   const {title, text, pageid, carousels} = question
+  // Create a seen set for glossary terms that will be shared across all Contents components
+  const [seenGlossaryTerms] = useState(() => new Set<string>())
```
Aprillion, line 158: please ask the coding assistant to fix this incorrect  react state usage - either use the setter function to create new sets on "change" (react state must be immutable) or use a mutable current value from `useRef` 🙏 

or we can discuss on discord during/after this weekend, I will be back home and can help more

Aprillion: also not sure we really need the double state - IMHO should be sufficient to have it only here and always pass as 2 props (the state and the setter function) to Contents child component

LeMurphant: Can't say I understand what is best practice here, but coding assistant suggested going the MutableRefObject route
https://github.com/StampyAI/stampy-ui/pull/994/commits/96ba465e62cbb1ae17a8f4e396780f49d22b85e6

app/components/Article/Contents.tsx
```diff
@@ -113,7 +114,7 @@ const glossaryInjecter = (pageid: string, glossary: Glossary, seen: Set<string>)
       const pattern = new RegExp(`(^|[^\\w-])(${normalizedAlias})($|[^\\w-])`, 'gi')
 
       result = result.replace(pattern, (match, prefix, matchedTerm, suffix) => {
-        const placeholder = `__GLOSSARY_${Math.floor(Math.random() * 100000)}__`
```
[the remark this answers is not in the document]
Aprillion: 🤞

app/components/Article/Contents.tsx
```diff
@@ -96,42 +96,57 @@ const normalizeForComparison = (text: string): string => {
...
+    let placeholderCounter = 0
+
+    // Replace ALL occurrences of each term with placeholders
+    // This prevents shorter terms from matching within longer terms
+    for (const {term, alias} of sortedGlossary) {
+      const normalizedAlias = normalizeForComparison(alias || '')
+      const pattern = new RegExp(`(^|[^\\w-])(${normalizedAlias})($|[^\\w-])`, 'gi')
+
+      result = result.replace(pattern, (match, prefix, matchedTerm, suffix) => {
+        const placeholder = `__GLOSSARY_${placeholderCounter++}__`
```
Aprillion, line 117: (minor) if there is only 1 loop and not doing replacements.push first and replacements.forEach second, then we wouldn't need even placeholder, placeholderCounter and just do the regex replace directly with the span (..but not a big deal if current code works fine)

Aprillion (approved):
please also check the merge conflict

jrhender (approved):
Looks good to me @LeMurphant !

PR #995, opened 2025-09-01
Fetch related question status

jrhender:
This PR filters out questions whose statuses that aren't `LIVE_ON_SITE` from the related questions that are included when fetching question details. Fixes: https://github.com/StampyAI/stampy-chat/issues/147
Approach initially discussed here: https://github.com/StampyAI/stampy-ui/pull/963#discussion_r2234016636

The approach taken is to fetch each related question individually and filter based on status. A new function `loadQuestionStatus` is introduced for this. `loadQuestionStatus` is cached for each individual question so that if it is related to multiple questions it can be reused.

Also considered calling `loadOnSiteAnswers` in `loadQuestionDetail` and then use that for filtering of the relatedQuestions. However, `loadOnSiteAnswers` might become unfeasible if more questions are added and would have some complexity needed to prevent recursion (as `loadOnSiteAnswers` loads questions that have their related questions).

Manual testing:
- [x] Navigate to a question page
- [x] Obtain human written answer to "What is a "warning shot"?" using chatbot -> the unpublished articles "Should we expect warning shots before an unrecoverable catastrophe?" and "Would warning shots" make a difference and, if so, would they be helpful or harmful?" should not displayed as suggestions
- [ ] Ask follow up to chatbot question 
- [ ] Click around the site some more

Possible TODOs:
- [ ] Test a bit more
- [ ] Look on Matomo to see if there metric we could look at to measure the latency impact of this change 
- [ ] filter on Status property rowId (less likely to be broken by row rename)
- [ ] decide if the `promise.all` usage is okay

### DIFF
--- a/app/server-utils/stampy.ts
+++ b/app/server-utils/stampy.ts
@@ -309,10 +309,13 @@ const extractJoined = (values: Entity[], mapper: Record<string, any>) =>
     .map((name) => (mapper && mapper[name]) || {name})
     .filter((i) => i)
 
-const convertToQuestion = async ({name, values, updatedAt} = {} as AnswersRow, request?: RequestForReload): Promise<Question> => {
+const convertToQuestion = async (
+  {name, values, updatedAt} = {} as AnswersRow,
+  request?: RequestForReload
+): Promise<Question> => {
   const markdown = extractText(values['Rich Text'])
   const {html, carousels} = renderText(extractText(values['UI ID']), values['Rich Text'])
-  
+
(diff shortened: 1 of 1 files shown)


jrhender:
@mruwnik thanks for the feedback! I'm open to the approach of having all of the questions fetched in order to fetch an individual question. I didn't use that approach because I figured it wouldn't scale as well if the number of questions grows. But I think that this scalability probably isn't a concern in the immediate future, so I will change this approach to use the "fetch all questions" approach.

Issue #1012, opened 2025-11-10
Markdown in chat answers is not rendered correctly


jrhender:
In https://github.com/StampyAI/stampy-chat/issues/25 , @LeMurphant reported that markdown formatting was appearing in Stampy Chat answers but was not being rendered correctly. This issue created to track this specific behaviour.

Example 1:
<img width="480" height="270" alt="Image" src="https://github.com/user-attachments/assets/1b7138f4-20b4-468f-a95b-ceda3c796114" />

Example 2:
<img width="807" height="644" alt="Image" src="https://github.com/user-attachments/assets/db3667e2-8a2e-4f6a-af1e-f974d1bc9740" />

LeMurphant:
John remarks that we already do something similar on the webpages with stampy-ui rendering Markdown from coda, we could reuse that code.

PR #1013, opened 2025-11-10
Render chatbot markdown

LeMurphant:
Fix https://github.com/StampyAI/stampy-ui/issues/1012

image

I also removed a bunch of mobile and popup options. Tested on mobile, does not seem to change anything, were they unused features? The popups didn't seem to work on the live website either.


app/components/Chatbot/ChatEntry.tsx
```diff
@@ -15,8 +15,6 @@ import StampyIcon from '~/components/icons-generated/Stampy'
 import PersonInCircleIcon from '~/components/icons-generated/PersonInCircle'
 import IconStampySmall from '~/components/icons-generated/StampySmall'
 import QuestionMarkIcon from '~/components/icons-generated/QuestionMark'
-import useIsMobile from '~/hooks/isMobile'
-import {togglePopup} from '../popups'
```
jrhender: It seems like this code might have been used in mobile views. I wasn't able to test the code and see the pop-up, but I think that in principle this code was in use so this is a change to the UX.

app/components/Chatbot/ChatEntry.tsx
```diff
@@ -119,57 +117,7 @@ const ReferenceSummary = ({
...
-        className={`reference-link ref-${index}`}
-        onClick={clickHandler}
-      >
-        <span>{index}</span>
-      </Link>
-      <ReferencePopup {...citation} className={shown ? 'shown' : 'hidden'} onClose={clickHandler} />
-    </span>
-  )
-}
+const md = new MarkdownIt({html: true, breaks: true})
```
jrhender:
```suggestion
const md = new MarkdownIt({html: false, breaks: true})
```

### Security Improvement: Use Placeholder-Based Citation Injection

The current approach injects citation HTML before markdown rendering with html: true, which means any malicious HTML/JavaScript in the LLM-generated content could be executed (XSS risk). Safer approach: Use placeholders to separate content escaping from citation injection.

### Suggested Implementation
```
1. Change MarkdownIt config (line 120):
const md = new MarkdownIt({html: false, breaks: true})
2. Replace the renderContentWithCitations function (lines 215-229) with:
  // Render content with markdown and inject citations
  const renderContentWithCitations = () => {
    if (!content) return null

    // Step 1: Replace citation markers with placeholders that markdown won't touch
    // This ensures markdown-it escapes any malicious content before we inject citations
    const placeholders = new Map<string, Citation>()
    let placeholderIndex = 0

    const contentWithPlaceholders = content.replace(/\[(\d+)\]/g, (match, refId) => {
      const ref = citationsMap?.get(refId)
      if (!ref) return match

      const placeholder = `|||CITATION${placeholderIndex}|||`
      placeholders.set(placeholder, ref)
      placeholderIndex++
      return placeholder
    })

    // Step 2: Render markdown safely (this escapes any HTML/XSS in the content)
    const renderedHtml = md.render(contentWithPlaceholders)

    // Step 3: Replace placeholders with actual citation HTML
    // Safe because: (a) content was escaped by markdown-it, (b) citation HTML is controlled by us
    let finalHtml = renderedHtml
    placeholders.forEach((ref, placeholder) => {
      const citationHtml = `<span class="ref-container"><a id="${ref.id}-${no}-ref" href="#${ref.id}-${no}" class="reference-link ref-${ref.index}"><span>${ref.index}</span></a></span>`
      finalHtml = finalHtml.replace(placeholder, citationHtml)
    })

    return <div dangerouslySetInnerHTML={{__html: finalHtml}} />
  }
```

### Why this is more secure:
✅ html: false prevents raw HTML injection from LLM responses
✅ Markdown-it escapes any malicious content before we inject citations
✅ Only our controlled citation HTML is injected post-processing
✅ Maintains all markdown formatting (bold, italic, lists, line breaks)
✅ Prevents XSS attacks from crafted LLM prompts

jrhender (commented):
There is a regression bug both in this PR and with the code I suggested where it takes two clicks to scroll down to the citation link. This isn't present on the master branch. https://www.loom.com/share/a80f3c775c074c3c894a75aa2bd84ffd
I think that it should be fixed before merging.

Claude proposed to me using the `Link` component as was done before, keeping everything in the React render tree. I wasn't able to get this to work though, so maybe a different approach would be better.

app/components/Chatbot/ChatEntry.tsx
LeMurphant: I also noticed this when it changed, but thought that given that we control the sources the risks were low. Claude agrees
```
  1. Low risk in practice - The content comes from your own LLM backend (stampy-chat), not user input
  2. Backend should sanitize - The LLM service should already be preventing malicious output
  3. LLMs are trained against this - Modern LLMs typically don't generate script tags or malicious HTML
```
Instead of changing the architecture, it suggested we use dompurify, which I tried in https://github.com/StampyAI/stampy-ui/pull/1013/commits/d900506aadb7eed81e5d095808d76e818882db9e

app/components/Chatbot/ChatEntry.tsx
jrhender: I wasn't able to get citation popups to work on mobile and if you aren't either then on balance I think it's okay to remove this code.

app/components/Chatbot/ChatEntry.tsx
jrhender: I agree it's low risk in our case. AI Safety.info doesn't have anything of value for an XSS attack. Still good to err on the side of caution though. I think the dompurify approach is good.

app/components/Chatbot/ChatEntry.tsx
```diff
@@ -258,13 +207,36 @@ const ChatbotReply = ({
...
+  // Render content with markdown and inject citations
+  const renderContentWithCitations = () => {
+    if (!content) return null
+
+    // Replace [N] citations with citation HTML (markdown-it will preserve this with html: true)
+    const processedContent = content.replace(/\[(\d+)\]/g, (match, refId) => {
+      const ref = citationsMap?.get(refId)
+      if (!ref) return match
+      const targetId = `${ref.id}-${no}`
+      return `<span class="ref-container"><a id="${targetId}-ref" href="#${targetId}" class="reference-link ref-${ref.index}" onclick="event.preventDefault(); document.getElementById('${targetId}')?.scrollIntoView({behavior: 'smooth', block: 'start'});"><span>${ref.index}</span></a></span>`
```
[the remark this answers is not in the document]
LeMurphant: It works for me in testing. Note that if you click on the reference before the text has finished displaying then it can't go there since it's not printed. Claude suggested rendering the references before the text, but I thought it was not worth it.

I'm ok converting Markdown to React, do you think this would be preferable to the current approach on principle?

jrhender (approved):

Issue #702, opened 2024-06-07
clicking footnotes leads to home page


melissasamworth:
Either the little number, or the arrows at the end of footnotes themselves

![Image](https://github.com/StampyAI/stampy-ui/assets/825910/bf48d078-1a2a-4732-aaef-e2cbf367838d)

![Image](https://github.com/StampyAI/stampy-ui/assets/825910/1bf67285-2b69-4b98-bd26-6cfe8bee83e6)

LeMurphant:
Unfortunately this still does not work as intended.
[footnotes.zip](https://github.com/user-attachments/files/16698016/footnotes.zip)

melissasamworth:
@LeMurphant seems done to me when I check! Could you check again in case I missed something, and then close out? (can you close out a task I created?)

LeMurphant:
@melissasamworth I'm still seeing the (wrong) behaviour I illustrated in the video in my previous comment. 

FYI yes I can close an issue even if I did not open it, but this might be because I have all the permissions.

melissasamworth:
@mruwnik could you work with @LeMurphant to address this?

LeMurphant:
@melissasamworth VLC media player should open the video, although I'm surprised that your default video player does not support it.

To reproduce it:
- find a page with a footnote, e.g. https://aisafety.info/questions/6182/What-are-the-potential-benefits-of-advanced-AI
- Click on a footnote number (e.g. 1)
- Sometimes, this goes to the footnote, sometimes not, but it reliably does so the second time, and the URL changes slightly
- When it goes to the footnote, click the "back" arrow. This reliably fails the first time and works the second time
Tested on the most recent version of Chrome

But yeah, that's probably either a nemo issue

Algon-33:
For some reason, the backlinks in the footnote now work in firefox on Windows 11. 
I tested it on this page: https://aisafety.info/questions/6953/Do-people-seriously-worry-about-existential-risk-from-AI 
Murphant tested this on Firefox too, and also found that they work. But on Chrome, the issue persists. Likewise for Microsoft Edge. Strangely, the first two clicks on footnotes in a page work. But then the issue appears. Oddly, it seems like on Edge the page will still move whenever you click on a backlink/footnote. It's just that it will instantly take you back to your previous position.

Issue #1014, opened 2025-11-17
Convert Markdown to React


LeMurphant:
See https://github.com/StampyAI/stampy-ui/pull/1013#discussion_r2529949748

Not a priority

PR #1019, opened 2025-11-24
Fix footnote backlinks in Chromium

LeMurphant:
Fix https://github.com/StampyAI/stampy-ui/issues/702

Tested in Firefox + Chrome. Also now has nice scrolling effect :)

I wanted to emulate solution we used in https://github.com/StampyAI/stampy-ui/pull/1013#issuecomment-3523223212, but could not because in the former case we generate the HTML as we go. I don't love that we are using different solutions to the same problem, @jrhender do you think we should standardize how we handle that?

https://github.com/StampyAI/stampy-ui/issues/1014 would be a more elegant solution, but is not happening for now.

### DIFF
--- a/app/components/Article/article.css
+++ b/app/components/Article/article.css
@@ -242,6 +242,8 @@ article .glossary-popup .button {
 
 article .link-popup .footnote {
   padding: var(--spacing-32) var(--spacing-32) var(--spacing-32) var(--spacing-32);
+  overflow-wrap: break-word;
+  word-break: break-word;
 }
 
 .contents .footnotes {


Algon-33 (commented):
Change smooth slide to instant movement

Algon-33:
@LeMurphant I can review this if you want me to.

LeMurphant:
> @LeMurphant I can review this if you want me to.

I'd rather try to address https://github.com/StampyAI/stampy-ui/issues/1020 before we merge. Furthermore, I'd like John to at least have a look since I am uncertain of the architectural aspect.

Issue #1022, opened 2025-12-04
Add heartbeat tracking


Algon-33:
On matomo, we currently can't track how long someone spent on an article if they don't take any actions after opening it. This is an issue for when we're doing surveys on site. We can resolve this by adding a "HeartBeatTimer" in the code, which pings the user's pc every 5 seconds to time how long they've been doing stuff. This will solve our analytics issue.

PR #1024, opened 2025-12-23
Heartbeat tracking fixed, and tested w/ mock Matomo for local dev

Algon-33:
Fixed the problems with the prior merge #1023 for issue #1022. The issue was the stray 5 in `window._paq || [5];` Also added a file, public/mock-matomo.js, to test the tracking code works correctly. This meant changes to the AnalticsTag in app/root.tsx, to allow for mock testing via an if statement. You can enable this in local dev by changing the MATOMO_DOMAIN in wrangler.toml to "{mock}" '.

### DIFF
--- a/app/root.tsx
+++ b/app/root.tsx
@@ -68,22 +68,22 @@ export const meta: MetaFunction<typeof loader> = ({data = {} as any}) => {
   const url = data.url ? new URL(data.url) : null
   const logo = url ? `${url.origin}/aisafety-logo.png` : '/aisafety-logo.png'
   return [
-    { title },
-    { name: 'description', content: description },
-    { property: 'og:url', content: data.url },
-    { property: 'og:type', content: 'article' },
-    { property: 'og:title', content: title },
-    { property: 'og:description', content: description },
-    { property: 'og:image', content: logo },
-    { property: 'og:image:type', content: 'image/png' },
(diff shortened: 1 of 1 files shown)


public/mock-matomo.js
```diff
@@ -0,0 +1,274 @@
+/**
+ * Mock Matomo Tracker for Local Development
+ *
+ * This simulates the Matomo analytics library to allow testing tracking code
+ * without needing the real Matomo server. All tracking calls are logged to the console.
```
jrhender: Would be nice to explain how this was written. I assume claude either looked at the SDK interface and wrote an implementation or just went off of it's inbuilt knowledge of Matomo. I think this is important as the testing is only useful in so far as this mock-matomo class matches the actual Matomo interface.

jrhender (approved):
> I talked it over with Claude, and we're pretty sure this shouldn't increase the bundle size for the user or result in a less performant site, because while the mock-matomo.js file is public, the code path requesting it can only be executed locally, not in production. So could you review the pull request, and if you're OK with it, merge it?

Makes sense to me 👍

public/mock-matomo.js
Algon-33: It's just a standard matomo implementation which matches the official API signatures, and Opus wrote the code by referencing official Matomo documentation.

PR #1047, opened 2026-04-02
refactor: tool-based RAG UI with block-based rendering

nerualauren:
> **Note**: This PR was assembled and reviewed by Claude Code. Lauren has reviewed the design and conversation ~~but has not yet read the full diff line-by-line.~~ [edit: have now read the full diff]

Companion PR to StampyAI/stampy-chat#204 -- updates the chat UI to support the new tool-based RAG pipeline.

## Summary

- **Block-based rendering**: thinking, tool_use+tool_result (merged into single collapsible), text blocks
- **Turn-based SSE protocol**: handles new `turn` events carrying tool results alongside existing `streaming`/`thinking` events
- **Citations from tool results**: extracted from `tool_result.ui_output` blocks instead of a separate `citations` SSE event
- **Cross-turn citation resolution**: `priorCitations` accumulated across conversation turns
- **Markdown + KaTeX**: `react-markdown` with `remark-math`/`rehype-katex` for math rendering
- **Memoized CitationMarkdown**: fixes perf issue where `withCitationProcessing` recreated component types every render, causing React unmount/remount flicker during streaming
- **Tool display**: friendly names ("Searched articles", "Viewed full article") in collapsible details with request/response
- **Model list**: updated to dateless Anthropic IDs (`claude-sonnet-4-6`, `claude-opus-4-6`, etc.)
- **CHATBOT_URL**: production default with automatic localhost detection for dev
- **Payload limit**: raised to 200KB for block-structured history with tool results

## Test plan

- [x] Local dev: chat works with tool calls visible in collapsible details
- [x] Citations render correctly with clickable reference links
- [x] Multi-turn conversations preserve citation numbering
- [x] Thinking blocks show animated ellipsis while active, "Thoughts" when done
- [x] Math expressions render via KaTeX
- [x] No flicker during streaming (CitationMarkdown memoization)
- [x] Production build hits `chat.stampy.ai`, local dev hits `localhost:3001`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/components/Chatbot/ChatEntry.tsx
+++ b/app/components/Chatbot/ChatEntry.tsx
@@ -1,12 +1,22 @@
-import {ComponentType, ReactNode, useRef, memo} from 'react'
+import {ComponentType, ReactNode, useRef, useMemo, memo, Fragment} from 'react'
 import {Link} from '@remix-run/react'
-import MarkdownIt from 'markdown-it'
-import DOMPurify from 'dompurify'
+import Markdown from 'react-markdown'
+import remarkMath from 'remark-math'
+import rehypeKatex from 'rehype-katex'
+import 'katex/dist/katex.min.css'
 import Contents from '~/components/Article/Contents'
 import Feedback, {logFeedback} from '~/components/Feedback'
 import useGlossaryInjection from '~/hooks/useGlossaryInjection'
 import './chat_entry.css'
-import type {Entry, AssistantEntry, StampyEntry, Citation, ErrorMessage} from '~/hooks/useChat'
+import {formatCitations} from '~/hooks/useChat'
+import type {
+  Entry,
+  AssistantEntry,
+  ContentBlock,
+  StampyEntry,
+  Citation,
+  ErrorMessage,
+} from '~/hooks/useChat'
 
 // icons
 import IconBotSmall from '~/components/icons-generated/BotSmall'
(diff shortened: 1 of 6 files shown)


app/components/Chatbot/ChatEntry.tsx
```diff
@@ -1,12 +1,22 @@
-import {ComponentType, ReactNode, useRef, memo} from 'react'
+import {ComponentType, ReactNode, useRef, useMemo, memo, Fragment} from 'react'
 import {Link} from '@remix-run/react'
-import MarkdownIt from 'markdown-it'
-import DOMPurify from 'dompurify'
+import Markdown from 'react-markdown'
```
nerualauren, line 3: switching to react-markdown for the chatbot for flexibility. also makes it reasonably easy to toss in katex (math rendering), which we want anyway.

package.json
```diff
@@ -15,14 +15,18 @@
     "copy-to-clipboard": "^3.3.3",
     "dompurify": "^3.3.0",
     "isbot": "^5.1.28",
+    "katex": "^0.16.23",
     "lodash": "^4.17.21",
     "markdown-it": "^14.1.0",
```
nerualauren, line 20: markdown-it isn't going away because it's used in articles. I poked claude several times about maybe could we just do one or the other, but the conclusion ended up being "well, maybe, but the bottleneck is human testing effort, and react-markdown handles more of what we need for chatbot ui".

LeMurphant (approved):

Algon-33:
Oh crud, for some reason I didn't merge this when talking to @nerualauren earlier! Anyway, I think lauren should merge this since apparently it needs to be done at the same time as another PR lauren has got otherwise things will break.

